PR #1, opened 2022-01-17
Add a GitHub action to run the package curation verification

mnonnenmacher:


.github/workflows/verification.yml
```diff
@@ -0,0 +1,29 @@
+name: Verification
+
+on:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
```
sschuberth, line 9: I believe this could be dropped as we'll never directly push to `main`.

sschuberth (changes requested):

.github/workflows/verification.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions/setup-java@v2
+        with:
+          distribution: adopt
+          java-version: 11
+          cache: gradle
+      - name: Verify package curations
+        uses: burrunan/gradle-cache-action@v1
+        with:
+          build-root-directory: tools/curation-generator
+          arguments: verifyPackageCurations
```
sschuberth, line 29: Only semi-related: Could you maybe add to https://github.com/oss-review-toolkit/ort-config/blob/main/tools/curation-generator/README.md that the tool is not only used to *generate* curations, but also to *verify* them, and maybe even explicitly document the `verifyPackageCurations` task?

.github/workflows/verification.yml
mnonnenmacher: Isn't this also triggered if a PR is merged? I thought it could be useful for the rare cases where two PRs pass individually but cause an issue when merged.

.github/workflows/verification.yml
mnonnenmacher: Good point, I will do this in a separate PR and also think about renaming the tool.

.github/workflows/verification.yml
mnonnenmacher: Yes, but for this repository I would keep it as the checks are a lot faster and broken curations would be very annoying for users.

sschuberth (approved):

PR #5, opened 2022-01-20
NuGet curations

mnonnenmacher:
Some curations for NuGet packages, see the commit messages for details.

### DIFF
--- a/curations/NuGet/_/Log4ALA.yml
+++ b/curations/NuGet/_/Log4ALA.yml
@@ -1,0 +1,5 @@
+- id: "NuGet::Log4ALA"
+  curations:
+    comment: "Set the VCS URL, found on the project homepage at https://ptv-logistics.github.io/Log4ALA/."
+    vcs:
+      url: "https://github.com/ptv-logistics/Log4ALA.git"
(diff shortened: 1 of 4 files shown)


curations/NuGet/_/Microsoft.NETCore.Platforms.yml
```diff
@@ -1,3 +1,8 @@
+- id: "NuGet::Microsoft.NETCore.Platforms:[5.0.0,5.0.13]"
+  curations:
+    comment: "Set the VCS path of the module inside the multi-module repository."
```
sschuberth, line 3: Nit: I realized too late that also the auto-generated path curations use a comment that just says that the path is set... which is obvious. IMO it would be nicer if the comment would mention *how* the path was concluded. So, auto-generated path curations should comment that some algorithm / heuristic was used, and manual curations like these should use a comment that makes clear that manual research was involved and the curation was created by a human.

curations/NuGet/_/Log4ALA.yml
```diff
@@ -0,0 +1,5 @@
+- id: "NuGet::Log4ALA"
+  curations:
+    comment: "Set the VCS URL."
```
sschuberth, line 3: Again, could you just briefly add how the correct URL was determined?

curations/NuGet/_/Microsoft.NETCore.Platforms.yml
mnonnenmacher: Fixed for the commits in this PR, let's talk about updating the generated curations offline. I have also removed one commit for which https://github.com/oss-review-toolkit/ort/pull/4979 should work.

sschuberth (approved):

PR #14, opened 2022-03-02
Declared license mappings

mnonnenmacher:
Allow and add several declared license mappings. See the commit messages for details.

### DIFF
--- a/curations/PyPI/_/asgiref.yml
+++ b/curations/PyPI/_/asgiref.yml
@@ -1,0 +1,7 @@
+- id: "PyPI::asgiref:(,4.0.0["
+  curations:
+    comment: |
+      The library is licensed under BSD-3-Clause, see:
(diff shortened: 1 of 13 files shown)


curations/PyPI/_/asgiref.yml
```diff
@@ -0,0 +1,7 @@
+- id: "PyPI::asgiref"
```
sschuberth, line 1: A general question: Should we be super-careful and limit the mapping to the range for the current major version (for example), to avoid the (highly theoretical) case that (in this example) the project relicenses itself from "BSD-3-Clause" to "BSD-2-Clause" but keeps the declared license at "BSD License"?

mnonnenmacher: I have added version ranges for up until to the current major version.

sschuberth (approved):

PR #15, opened 2022-03-11
package-configurations: Add various Composer packages

tsteenbe:


package-configurations/Composer/phpunit/php-code-coverage/9.2.10/vcs.yml
```diff
@@ -0,0 +1,13 @@
+---
+id: "Composer:phpunit:php-code-coverage:9.2.10"
+vcs:
+  type: "Git"
+  url: "https://github.com/sebastianbergmann/php-code-coverage.git"
+  revision: "d5850aaf931743067f4bfc1ae4cbd06468400687"
```
sschuberth, line 6: Reading this reminds me of something I believe we discussed at some point, but didn't conclude anything IIRC: Should we use tag names (if existing) instead of hard-coding the full revision SHA1?

Because
- it's more error prone for the creator to insert the correct revision
- it's more cumbersome for the reviewer to verify that the revision matches the version

I know we argued at some point that tags could be deleted an recreated, but should we really thing this far?

sschuberth: And another thing that just struck me: There is no top-level comment field to explain how the revision to match was determined.

fviernau: > And another thing that just struck me: There is no top-level comment field to explain how the revision to match was determined.

The revision should always be the one which has been determined by the downloader automatically. Just to be sure, the revision does not assign a revision to VCS info, but it is matching it against the revision of the scan result.

Why do you think a comment is needed?

sschuberth: > Just to be sure, the revision does not assign a revision to VCS info, but it is matching it against the revision of the scan result.

Yes, that is also something that struck me in the past, and that we should make more clear IMO.

Esp. when coming from package curation files (where only the `id` is used for matching and all other content overrides something), reading package configurartion files as a human being can be confusing, because it's totally unclear from the file structure alone which parts are used for matching, and which parts are used for overriding information.

IMO it would help to nest the matching `sourceArtifactUrl` and `vcs` fields under something like a `match:` field, and nest the `pathExcludes` and `licenseFindingCurations` fields under something like an `apply:` field.

> Why do you think a comment is needed?

With the above-mentioned nesting, a comment would maybe not be needed anymore for readability.

fviernau: Another point I forgot: In order to apply package curations, the revision needs to be matched against the resolved revision of scan results. So, it cannot be done 'offline' anymore, in case its not a SHA1. So, it becomes slow.

sschuberth: > So, it cannot be done 'offline' anymore, in case its not a SHA1.

Good point, a pure string comparison would not suffice.

sschuberth: > Further on, the revision can be extracted from the OrtResult given the identifier automatically (e.g. ListLicensesCommand). So I believe it's not error prone for the creator and no added effort.

Yeah, we should probably promoted the _helper-cli_ more in that context, or yet better, add the workflow to create package configurations based on OrtResults to the ORT Workbench as a GUI operation 😁

sschuberth: > why it's combersome is that we currently don't have a way to share scan results with the public.

Ping @pombredanne as a reminder that we need to tackle this 😉

sschuberth: > > why it's combersome is that we currently don't have a way to share scan results with the public.
> 
> Ping @pombredanne as a reminder that we need to tackle this 😉

I just came across https://dataset.at/, which looks interesting for sharing data!

fviernau (approved):

PR #16, opened 2022-03-11
package-configurations: Add various Crate packages

tsteenbe:


package-configurations/Crate/_/regex/1.2.1/vcs.yml
```diff
@@ -0,0 +1,23 @@
...
+path_excludes:
+  - pattern: "**/*test*/**"
+    reason: "TEST_OF"
+  - pattern: "**/examples/**"
+    reason: "EXAMPLE_OF"
+  - pattern: "bench/**"
+    reason: "TEST_OF"
+  - pattern: "ci/**"
+    reason: "BUILD_TOOL_OF"
+  - pattern: "examples/**"
```
fviernau, line 16: This IMO should be redundant with the one in line 10, but IIRC it isn't due to how the implementation works currently. Shall we have a ticket for this in ORT?

fviernau (approved):

package-configurations/Crate/_/regex/1.2.1/vcs.yml
tsteenbe: @mnonnenmacher  @fviernau  Are you Ok to merge this PR and then we update this once https://github.com/oss-review-toolkit/ort/issues/4357 is done?

PR #18, opened 2022-03-18
curations: Add VCS curations for poi-ooxml-schemas

MarcelBochtler:


curations/Maven/org.apache.poi/poi-ooxml-schemas.yml
```diff
@@ -0,0 +1,45 @@
+- id: "Maven:org.apache.poi:poi-ooxml-schemas:3.17"
+  curations:
+    comment: |
+      No source artifact is published and no SCM tag is in the POM, but the repository was found via Internet search.
+    vcs:
+      type: "Git"
+      url: "https://github.com/apache/poi.git"
+      revision: "REL_3_17_FINAL"
```
sschuberth, line 8: Do we really need version-specific curations? I believe ORT should be able to find these tags...

fviernau (changes requested):

fviernau (approved):

PR #19, opened 2022-03-24
Add declared license mappings for atlassian packages

MarcelBochtler:


curations/Maven/com.atlassian.event/atlassian-event.yml
```diff
@@ -0,0 +1,5 @@
+- id: "Maven:com.atlassian.event:atlassian-event:[3.0.0,6.0.0["
```
sschuberth, line 1: Commit message:
- s/Curations/curations/
- Could we use the tag name in the permalink for radability, like https://bitbucket.org/atlassian/atlassian-event/src/atlassian-event-5.0.0/pom.xml#lines-21:22? (Also for the next commit.)

sschuberth (changes requested):

sschuberth (dismissed):

curations/Maven/com.atlassian.event/atlassian-event.yml
fviernau, line 1: What's the rationale for the version range?
Can we accept the risk of making the assumption that the term "BSD License" will always refer to "BSD-3-Clause" for this library? So, drop the version?

curations/Maven/com.atlassian.event/atlassian-event.yml
```diff
@@ -0,0 +1,5 @@
...
+  curations:
+    comment: "Declared license links to the BSD-3-Clause license."
```
fviernau, line 3: I'd tend to add the the link to the "BSD License" term in the POM file to the comment. Would it make sense to make this a rule for declared license mapping curations?

fviernau (changes requested):

curations/Maven/com.atlassian.event/atlassian-event.yml
sschuberth: > Can we accept the risk

That was exactly the point of an internal discussion, and we concluded "no". See also the discussion in https://github.com/oss-review-toolkit/ort-config/pull/14.

fviernau: > That was exactly the point of an internal discussion, and we concluded "no". See also the discussion in #14.

Thanks! Did this internal discussion cover also how this curation has to be reviewed? Does the reviewer need to check manually all releases from 3.0.0 until 6.0.0 whether all the versions in that range have used "BSD-3-Clause"?

sschuberth: IMO the *reviewer* of the PR should not review in that detail, esp. if the contribution comes from a trusted contributor. However, the *contributor* IMO should have at least checked the bounds of the version range.

fviernau: So, the rule seems to be:

1. Only checking the bounds is ok (two versions would be checked)
2. Dropping the version range completely is not ok (just one version would be checked)

So, checking 2 versions out of N is ok and checking 1 version out of M is not ok, with N >> 2 and M >> 1.
I believe these are quite similar in terms of risk. I'd challenge that and vote for a bit more risk tolerance in this
particular case. Any thoughts on this? 

I don't wanna duplicate any already done discussion, but I didn't find much under discussion in #14 hinted to above.

sschuberth: > with N >> 2 and M >> 1

The point is that M >> N (M is much bigger than N). So yes, I believe that checking two versions in a rather small range is (much) better than checking one version in a basically infinite range.

Plus, the hope is that many libraries would apply semantic versioning, and bump the major version on a license change. So limiting the upper bound to the next major version makes sense IMO.

fviernau: I see, this convinces me that it makes a non-neglegtible difference in terms of risk.
So, I agree with your decision.

curations/Maven/com.atlassian.event/atlassian-event.yml
MarcelBochtler: I added the URL to the pom.xml for the latest tag in the comment.

fviernau (approved):

curations/Maven/com.atlassian.event/atlassian-event.yml
fviernau: Thanks!

PR #22, opened 2022-04-12
SW360 VCS curation for http-support

MarcelBochtler:


curations/Maven/org.eclipse.sw360/http-support.yml
```diff
@@ -0,0 +1,8 @@
+- id: "Maven:org.eclipse.sw360:http-support"
+  curations:
+    comment: |
+      No SCM tag is in the POM.
+    vcs:
+      type: "Git"
+      url: "https://github.com/eclipse/sw360.git"
```
sschuberth, line 7: Could you also contribute these fixes to their POMs straight to SW360?

curations/Maven/org.eclipse.sw360/http-support.yml
```diff
@@ -0,0 +1,8 @@
+- id: "Maven:org.eclipse.sw360:http-support"
+  curations:
+    comment: |
+      No SCM tag is in the POM.
```
sschuberth, line 4: Sorry for not mentioning this earlier, but we should also say that no sources artifact is published either. Otherwise we'd not need this.

curations/Maven/org.eclipse.sw360/http-support.yml
sschuberth: But the published POM does not seem to refer to a parent: https://repo.eclipse.org/content/groups/releases/org/eclipse/sw360/http-support/13.2.0-28/http-support-13.2.0-28.pom

sschuberth (approved):

PR #27, opened 2022-04-21
curations: Add VCS and metadata curation for netty's boringssl

MarcelBochtler:


curations/Maven/io.netty/netty-tcnative-boringssl-static.yml
```diff
@@ -0,0 +1,9 @@
+- id: "Maven:io.netty:netty-tcnative-boringssl-static"
+  curations:
+    comment: |
+      No SCM tag is declared in the POM for this metadata package.
```
sschuberth, line 4: Actually the SCM tag *is* [declared in the POM](https://repo1.maven.org/maven2/io/netty/netty-tcnative-boringssl-static/2.0.51.Final/netty-tcnative-boringssl-static-2.0.51.Final.pom), but with a wrong URL path. That's a typical problem with Maven packages, and I wanted to discuss with you and @mnonnenmacher a more generic solution for that.

sschuberth (changes requested):

curations/Maven/io.netty/netty-tcnative-boringssl-static.yml
sschuberth: See https://github.com/oss-review-toolkit/ort/pull/5272.

curations/Maven/io.netty/netty-tcnative-boringssl-static.yml
```diff
@@ -0,0 +1,9 @@
...
+    is_meta_data_only: true
```
sschuberth, line 5: This bit, however, would still be needed even with https://github.com/oss-review-toolkit/ort/pull/5272.

sschuberth (approved):

PR #28, opened 2022-04-21
curations: Add declared license mapping for antlr4

MarcelBochtler:


curations/Maven/org.antlr/antlr4-runtime.yml
```diff
@@ -0,0 +1,7 @@
+- id: "Maven:org.antlr:antlr4-runtime:(,5.0.0["
```
sschuberth, line 1: It would make me feel better to start the range at version 4.0 (inclusive), even if the artifact name already contains the major version. Same below.

curations/Maven/org.antlr/antlr4-runtime.yml
```diff
@@ -0,0 +1,7 @@
...
+  curations:
+    comment: |
+      The library is licensed under BSD-3-Clause, see:
+      https://github.com/antlr/antlr4/blob/master/LICENSE.txt
```
sschuberth, line 5: Please use a permalink, probably using the tag for version 4.0.

sschuberth (changes requested):

sschuberth (approved):

PR #30, opened 2022-04-22
curations: Add VCS URLs for log4net

sschuberth:
Signed-off-by: Sebastian Schuberth <sebastian.schuberth@bosch.io>

### DIFF
--- a/curations/NuGet/_/log4net.yml
+++ b/curations/NuGet/_/log4net.yml
@@ -2,6 +2,8 @@
   curations:
     comment: "Set the VCS revision because there are multiple tags matching the version."
     vcs:
+      type: Git
+      url: https://github.com/apache/logging-log4net.git
       revision: "rel/2.0.8"
 - id: "NuGet::log4net:2.0.11"
   curations:
--- a/curations/NuGet/_/log4net.yml
+++ b/curations/NuGet/_/log4net.yml
@@ -5,6 +5,18 @@
       type: Git
       url: https://github.com/apache/logging-log4net.git
       revision: "rel/2.0.8"
+- id: "NuGet::log4net:2.0.9"
+  curations:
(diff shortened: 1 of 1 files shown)


curations/NuGet/_/log4net.yml
```diff
@@ -2,19 +2,27 @@
...
     vcs:
+      type: Git
+      url: https://github.com/apache/logging-log4net.git
       revision: "rel/2.0.8"
 - id: "NuGet::log4net:2.0.11"
   curations:
     comment: "Set the VCS revision because there are multiple tags matching the version."
     vcs:
+      type: Git
+      url: https://github.com/apache/logging-log4net.git
```
mnonnenmacher, line 13: Starting with 2.0.10 repository metadata should be available, compare:
https://nuget.info/packages/log4net/2.0.9
https://nuget.info/packages/log4net/2.0.10

mnonnenmacher (changes requested):

mnonnenmacher:
Since your are at it, would you mind adding a revision curation for 2.0.14 for which also an RC tag exists?

curations/NuGet/_/log4net.yml
sschuberth: Ah, that nuget.info site is very useful, thanks!

sschuberth:
> Since your are at it, would you mind adding a revision curation for 2.0.14 for which also an RC tag exists?

Do you think we should hard-code some heuristic regarding tag prefixes to work around this?

mnonnenmacher:
> > Since your are at it, would you mind adding a revision curation for 2.0.14 for which also an RC tag exists?
> 
> Do you think we should hard-code some heuristic regarding tag prefixes to work around this?

If you have a good idea how to improve go for it, but I'm a bit hesitant to touch that code.

mnonnenmacher (approved):

PR #33, opened 2022-05-02
Tools: Move code to buildSrc

mnonnenmacher:
See the commit messages for details.


tools/curations/build.gradle.kts
```diff
@@ -1,331 +1,16 @@
...
-val curationsDir = rootDir.resolve("../../curations")
-val mapper = YAMLMapper(YAMLFactory().disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER)).apply(mapperConfig)
+import org.ossreviewtoolkit.tools.curations.GenerateAspNetCoreCurationsTask
+import org.ossreviewtoolkit.tools.curations.GenerateAzureSdkForNetCurationsTask
+import org.ossreviewtoolkit.tools.curations.GenerateDotNetRuntimeCurationsTask
+import org.ossreviewtoolkit.tools.curations.VerifyPackageCurationsTask
 
 buildscript {
     repositories {
         mavenCentral()
```
sschuberth, line 8: Why is this still needed in here?

tools/curations/build.gradle.kts
```diff
@@ -1,331 +1,16 @@
...
-val versionRegex = Regex("^v?\\d+\\.\\d+\\.\\d+\$")
-
-fun String.isVersion() = matches(versionRegex)
+tasks.register("generateAspNetCoreCurations", GenerateAspNetCoreCurationsTask::class)
+tasks.register("generateAzureSdkForNetCurations", GenerateAzureSdkForNetCurationsTask::class)
+tasks.register("generateDotNetRuntimeCurations", GenerateDotNetRuntimeCurationsTask::class)
 
-fun Identifier.toCurationPath() =
-    "${type.encodeOr("_")}/${namespace.encodeOr("_")}/${name.encodeOr("_")}.yml"
+tasks.register("verifyPackageCurations", VerifyPackageCurationsTask::class)
```
sschuberth, line 16: More compact:

```
tasks {
    register<GenerateAspNetCoreCurationsTask>("generateAspNetCoreCurations")
    register<GenerateAzureSdkForNetCurationsTask>("generateAzureSdkForNetCurations")
    register<GenerateDotNetRuntimeCurationsTask>("generateDotNetRuntimeCurations")
    register<VerifyPackageCurationsTask>("verifyPackageCurations")
}
```

tools/curations/buildSrc/src/main/kotlin/AbstractGenerateCurationsTask.kt
```diff
@@ -0,0 +1,110 @@
...
+    init {
+        group = "generate curations"
+    }
+
+    fun getFilesFromRepository(
+        owner: String,
+        repository: String,
+        ref: String = "main"
+    ): List<String> {
+        val client = OkHttpClient()
```
sschuberth, line 36: Ho about moving this to a class variable to share the `client` across functions?

tools/curations/buildSrc/src/main/kotlin/AbstractGenerateCurationsTask.kt
```diff
@@ -0,0 +1,110 @@
...
+
+        val request =
+            Request.Builder()
+                .url("https://api.github.com/repos/$owner/$repository/git/trees/$ref?recursive=1")
+                .header(
+                    "Authorization",
+                    okhttp3.Credentials.basic(githubUsername, githubToken)
+                )
+                .build()
```
sschuberth, line 45: You could also move this common code out to a prepared builder on the class level, and refine that builder here with the URL, like `githubRequest.newBuilder().url().build()`.

sschuberth (changes requested):

tools/curations/build.gradle.kts
mnonnenmacher: It isn't.

tools/curations/build.gradle.kts
```diff
@@ -1,331 +1,10 @@
...
-import org.ossreviewtoolkit.utils.common.safeMkdirs
-import org.ossreviewtoolkit.utils.spdx.SpdxExpression.Strictness
+import org.ossreviewtoolkit.tools.curations.GenerateAspNetCoreCurationsTask
+import org.ossreviewtoolkit.tools.curations.GenerateAzureSdkForNetCurationsTask
+import org.ossreviewtoolkit.tools.curations.GenerateDotNetRuntimeCurationsTask
+import org.ossreviewtoolkit.tools.curations.VerifyPackageCurationsTask
 
-val githubUsername: String by project
-val githubToken: String by project
+tasks.register<GenerateAspNetCoreCurationsTask>("generateAspNetCoreCurations")
```
sschuberth, line 6: Just double checking, do you prefer to repeat `task.` instead of wrapping in `tasks {}`?

tools/curations/buildSrc/src/main/kotlin/AbstractGenerateCurationsTask.kt
```diff
@@ -0,0 +1,101 @@
...
+        YAMLMapper(YAMLFactory().disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER)).apply(mapperConfig)
+
+    private val httpClient = OkHttpClient()
+
+    init {
+        group = "generate curations"
+    }
+
+    private fun gitHubRequestBuilder() =
+        Request.Builder().header("Authorization", okhttp3.Credentials.basic(githubUsername, githubToken))
```
sschuberth, line 34: This could be moved to a property to avoid recreation on each function call. However, that requires to call `build()` here, and `newBuilder()` in the caller. Which then effectively is also something like doing this twice... your choice.

sschuberth: Also, maybe import `okhttp3.Credentials` instead of using a FQN.

mnonnenmacher: > This could be moved to a property to avoid recreation on each function call. However, that requires to call `build()` here, and `newBuilder()` in the caller. Which then effectively is also something like doing this twice... your choice.

I'm not quite sure how you imagine this to look like, but if it's about a performance optimization to not create a new builder for each call I think it's irrelevant because these functions are not called often.

sschuberth: For reference, this could look like:

    private val gitHubRequest = Request.Builder()
        .header("Authorization", Credentials.basic(githubUsername, githubToken)).build()

And then on the call side:

    val request = gitHubRequest.newBuilder()
        .url("https://api.github.com/repos/$owner/$repository/git/trees/$ref?recursive=1")
        .build()

sschuberth (approved):

PR #34, opened 2022-05-04
Add verification for package configurations

mnonnenmacher:
See the commit messages for details.

### DIFF
--- a/tools/curations/build.gradle.kts
+++ b/tools/curations/build.gradle.kts
@@ -11,4 +11,10 @@ tasks {
 
     register<VerifyPackageConfigurationsTask>("verifyPackageConfigurations")
     register<VerifyPackageCurationsTask>("verifyPackageCurations")
+
+    register("verify") {
+        group = "verification"
+
+        setDependsOn(this@tasks.filter { it.group == "verification" && it != this })
+    }
 }
--- a/tools/curations/build.gradle.kts
+++ b/tools/curations/build.gradle.kts
@@ -1,6 +1,7 @@
 import org.ossreviewtoolkit.tools.curations.GenerateAspNetCoreCurationsTask
 import org.ossreviewtoolkit.tools.curations.GenerateAzureSdkForNetCurationsTask
 import org.ossreviewtoolkit.tools.curations.GenerateDotNetRuntimeCurationsTask
+import org.ossreviewtoolkit.tools.curations.VerifyPackageConfigurationsTask
 import org.ossreviewtoolkit.tools.curations.VerifyPackageCurationsTask
 
 tasks {
@@ -8,5 +9,6 @@ tasks {
     register<GenerateAzureSdkForNetCurationsTask>("generateAzureSdkForNetCurations")
     register<GenerateDotNetRuntimeCurationsTask>("generateDotNetRuntimeCurations")
 
+    register<VerifyPackageConfigurationsTask>("verifyPackageConfigurations")
     register<VerifyPackageCurationsTask>("verifyPackageCurations")
 }
--- a/tools/curations/buildSrc/src/main/kotlin/VerifyPackageConfigurationsTask.kt
+++ b/tools/curations/buildSrc/src/main/kotlin/VerifyPackageConfigurationsTask.kt
@@ -1,0 +1,86 @@
+package org.ossreviewtoolkit.tools.curations
+
+import com.fasterxml.jackson.module.kotlin.readValue
+import org.gradle.api.DefaultTask
+import org.gradle.api.GradleException
+import org.gradle.api.tasks.TaskAction
+import org.ossreviewtoolkit.model.config.PackageConfiguration
+import org.ossreviewtoolkit.model.mapper
+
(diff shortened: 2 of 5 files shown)


tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
```diff
@@ -16,10 +16,13 @@ abstract class VerifyPackageCurationsTask : DefaultTask() {
 
     @TaskAction
     fun verify() {
-        var count = 0
+        var fileCount = 0
+        var curationCount = 0
         val issues = mutableListOf<String>()
 
         curationsDir.walk().filter { it.isFile }.forEach { file ->
+            fileCount++
```
sschuberth, line 24: If you did something like `val curationsFiles = curationsDir.walk().filter { it.isFile }` then you could directly assign `val fileCount = curationsFiles.size`.

tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
```diff
@@ -77,17 +82,18 @@ abstract class VerifyPackageCurationsTask : DefaultTask() {
...
             }.onFailure { e ->
                 issues += "Could not parse curations from file '$relativePath': ${e.message}"
             }
         }
 
         if (issues.isNotEmpty()) {
-            throw GradleException("Found ${issues.size} curation issues:\n${issues.joinToString("\n")}")
+            throw GradleException(
+                "Found ${issues.size} curation issues in $fileCount package curation files containing $curationCount " +
+                        "curations:\n${issues.joinToString("\n")}"
```
sschuberth, line 93: Nit: While at it maybe use "(s)" syntax to match both singular and plural.

tools/curations/buildSrc/src/main/kotlin/VerifyPackageConfigurationsTask.kt
```diff
@@ -0,0 +1,86 @@
...
+        group = "verification"
+    }
+
+    @TaskAction
+    fun verify() {
+        var fileCount = 0
+        val issues = mutableListOf<String>()
+
+        packageConfigurationsDir.walk().filter { it.isFile }.forEach { file ->
+            fileCount++
```
sschuberth, line 21: Also here get `fileCount` from the size of the intermediate collection.

tools/curations/buildSrc/src/main/kotlin/VerifyPackageConfigurationsTask.kt
```diff
@@ -0,0 +1,86 @@
...
+                            "package name."
+                }
+
+                if (config.id.version.isBlank()) {
+                    issues += "Only package configurations for specific versions are allowed, but the configuration " +
+                            "for package '${config.id.toCoordinates()}' in file '$relativePath' does not have a " +
+                            "version."
+                }
+
+                config.sourceArtifactUrl?.let { sourceArtifact ->
```
sschuberth, line 44: Nit: `let` is usually used when the return value of the lambda matters. Maybe `also` or `run` are a better fit? Similar below.

tools/curations/build.gradle.kts
```diff
@@ -11,4 +11,10 @@ tasks {
 
     register<VerifyPackageConfigurationsTask>("verifyPackageConfigurations")
     register<VerifyPackageCurationsTask>("verifyPackageCurations")
+
+    register("verify") {
+        group = "verification"
+
+        setDependsOn(this@tasks.filter { it.group == "verification" && it != this })
```
sschuberth, line 18: Can we drop the `this@` here? And is the `it != this` condition really required, as I'd assume the task is only present in `tasks` *after* this registration?

sschuberth (changes requested):

tools/curations/build.gradle.kts
mnonnenmacher: `this@tasks` is required because `this` points to the current task. And I expected the same, but without `it != this` Gradle complains about a circular task dependency.

tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
mnonnenmacher: This is a sequence which does not have a `size` property.

tools/curations/buildSrc/src/main/kotlin/VerifyPackageConfigurationsTask.kt
mnonnenmacher: See related comment above.

tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
mnonnenmacher: Ok, then I'll keep it as is.

sschuberth (approved):

PR #38, opened 2022-06-20
Fix a compile issue with policy rules and add initial versions of further configuration files

fviernau:


evaluator-rules/src/main/resources/example.rules.kts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019-2022 HERE Europe B.V.
+ * Copyright (C) 2019 HERE Europe B.V.
```
tsteenbe, line 2: This should be "Copyright (C) 2019-2022 HERE Europe B.V."

tsteenbe (changes requested):

PR #41, opened 2022-06-28
Examplify how to define custom license ids / texts

fviernau:
ScanCode version '30.0.1' does not detect any license in [1]. So, add a
custom license for the time being, until ScanCode becomes capable of
detecting [1]. The text was copied from [1] and formatted via [2].

[1] https://github.com/gatling/gatling-highcharts/blob/aca5f92f517d58c33c38fccc10649a75b3b72325/LICENSE
[2] `fmt -w  80 < LicenseRef-ort-GatlingCorp-HighCharts`


mnonnenmacher (dismissed):
To complete the example, could you also add a related curation that makes use of the custom license reference?

tsteenbe (changes requested):

tsteenbe (approved):

PR #42, opened 2022-06-29
curations: Add metadata only curation for aws-java-sdk

MarcelBochtler:
See https://github.com/aws/aws-sdk-java.

### DIFF
--- a/curations/Maven/com.amazonaws/aws-java-sdk.yml
+++ b/curations/Maven/com.amazonaws/aws-java-sdk.yml
@@ -1,0 +1,6 @@
+- id: "Maven:com.amazonaws:aws-java-sdk:[1.9.0,)"
+  curations:
+    comment: |
+      From version 1.9.0 on this package is the parent POM for multiple packages in the AWS SDK for Java. Previous
+      versions contain class files and distribute a *-sources.jar.
+    is_meta_data_only: true


curations/Maven/com.amazonaws/aws-java-sdk.yml
```diff
@@ -0,0 +1,5 @@
+- id: "Maven:com.amazonaws:aws-java-sdk"
```
sschuberth, line 1: Looks like this hasn't always been a parent POM only. Going back as far as https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk/1.0.000 shows that the [JAR](https://repo1.maven.org/maven2/com/amazonaws/aws-java-sdk/1.0.000/aws-java-sdk-1.0.000.jar) is actually 1.3 MiB in size and contains classes. So we probably need a lower-bound for the version here.

sschuberth (changes requested):

curations/Maven/com.amazonaws/aws-java-sdk.yml
```diff
@@ -0,0 +1,6 @@
+- id: "Maven:com.amazonaws:aws-java-sdk:[1.9.0,)"
+  curations:
+    comment: |
+      From version 1.9.0 on this package is the parent POM for multiple packages in the AWS SDK for Java. Previous
+      versions contain source code and distribute a *-sources.jar.
```
sschuberth, line 5: Nit: "contain class files"

sschuberth (changes requested):

sschuberth (approved):

PR #56, opened 2022-09-19
Minor rules fixes

sschuberth:
Please have a look at the individual commit messages for the details.


evaluator-rules/src/main/resources/example.rules.kts
```diff
@@ -1103,7 +1103,7 @@ fun RuleSet.dependencyInProjectSourceRule() = projectSourceRule("DEPENDENCY_IN_P
 
 fun RuleSet.deprecatedScopeExludeInOrtYmlRule() = ortResultRule("DEPRECATED_SCOPE_EXCLUDE_REASON_IN_ORT_YML") {
     val reasons = ortResult.repository.config.excludes.scopes.mapTo(mutableSetOf()) { it.reason }
-    val deprecatedReasons = setOf(ScopeExcludeReason.TEST_TOOL_OF)
+    val deprecatedReasons = setOf(ScopeExcludeReason.TEST_DEPENDENCY_OF)
```
fviernau, line 1106: Please provide a rationale which this commit is done.

This is not obvious at all as the rule is checking for use of deprecated reasons.
So the reasons in the list must be deprecated.

fviernau (changes requested):

evaluator-rules/src/main/resources/example.rules.kts
sschuberth: Ah, I was misreading the intention of the rule. I'm now suppressing the deprecation like in https://github.com/oss-review-toolkit/ort/commit/8aeb944ecc2139ea51f1b45f7e19fc1596e4e0de.

evaluator-rules/src/main/resources/example.rules.kts
```diff
@@ -1062,7 +1062,7 @@ fun RuleSet.copyleftLimitedInSourceRule() = packageRule("COPYLEFT_LIMITED_IN_SOU
 
         val licenseSourceName = licenseSource.name.lowercase()
         val message = if (licenseSource == LicenseSource.DETECTED) {
-            if (pkg.id.type == "Unmanaged") {
+            if (isType("Unmanaged")) {
```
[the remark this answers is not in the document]
sschuberth: > Hm, there are a couple of other case distinction which still use `pkg.id.type`.

True. I could change those `when`s, too.

> Personally, I don't like the readability of `isType()` as it is not obvious anymore that type refers to `package.type`

Well, `isType()` is in the scope of `PackageRule`, so it's not that unclear I'd say.

> Code is identical isn't it?

Indeed. I'll drop it in another commit.

fviernau (approved):

Issue #59, opened 2022-09-26
Add a Gradle task for generating a `license-classifications.yml` from ScanCode's license data


fviernau:
Make it easy to use ScanCode's license categorization in ORT.

This can make use of https://scancode-licensedb.aboutcode.org/index.json.

sschuberth:
Just FYI, there's already @maxhbr's somewhat similar generated https://github.com/maxhbr/LDBcollector/blob/generated/ort/license-classifications.yml for ORT.

maxhbr:
the code in https://github.com/maxhbr/LDBcollector/blob/generated/ort/license-classifications.yml contains the classification from ScanCode and other sources. (Just restricting LDBcollector to ScanCode could even produce what you look for.)

fviernau:
Follow-up: #66.

PR #60, opened 2022-09-28
tools/curations: Make several Gradle tasks non-abstract

fviernau:
These tasks don't need to be abstract as they don't have any abstract functions. So, drop the `abstract` keyword in favor of `open`, as Gradle tasks are required to be non-final.

Part of #59.

### DIFF
--- a/license-classifications.yml
+++ b/license-classifications.yml
@@ -1,364 +1,10700 @@
+#
+# This ORT configuration file is provided as an example only. It
+# demonstrates the general configuration capabilities of ORT and does not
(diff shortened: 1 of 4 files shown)


sschuberth:
> These tasks don't need to be abstract as they don't have any abstract functions. So, drop the `abstract` keyword in favor of `open`, as Gradle tasks are required to be non-final.

FYI, this text and the PR title do not seem to match what's being done in the commits.

fviernau:
> FYI, this text and the PR title do not seem to match what's being done in the commits.

This was indeed wrong, I've closed this in favor of #62.

PR #61, opened 2022-09-28
license-classification: Drop the redundant 'strong-copyleft' category

fviernau:
The categories 'copyleft' and 'strong-copyleft' have identical semantics, while 'copyleft' has more usages and aligns with ScanCode's category name. So, align on using 'copyleft' as category for "Strong Copyleft" licenses.

Part of #59.

### DIFF
--- a/license-classifications.yml
+++ b/license-classifications.yml
@@ -11,7 +11,6 @@
 
 categories:
 - name: "copyleft"
-- name: "strong-copyleft"
 - name: "copyleft-limited"
 - name: "permissive"
   description: "Licenses with permissive obligations."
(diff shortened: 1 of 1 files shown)


sschuberth:
AFAIK, the most common terms for the "big three" license categories are "strong copyleft", "weak copyleft", and "permissive". While I agree that here "copyleft" and "strong copyleft" have identical semantics, I would have preferred to align on the more common "strong copyleft" term, and while at it also rename "copyleft-limited" to "weak-copyleft" (assuming that's what's mean with "copyleft-limited").

tsteenbe (approved):

sschuberth:
> Given that, do you agree?

I guess it's a bit weird for ScanCode to use those names; but yes, for the sake of easily generating the classifications, it make sense to use whatever the source of choice uses.

PR #62, opened 2022-09-28
tools/curations: Make several Gradle tasks non-abstract

fviernau:
These tasks don't need to be abstract as they don't have any abstract functions. So, drop the `abstract` keyword in favor of `open`, as Gradle tasks are required to be non-final.


sschuberth:
Looks like `AbstractGenerateCurationsTask` itself can be just `open`, too. Would you mind doing that change, and while at it also rename the class to not contain "Abstract" anymore?

sschuberth (approved):

PR #64, opened 2022-09-29
Switch to a generated `license-classifications.yml` based on ScanCode's License DB

### LINKED ISSUE #59: Add a Gradle task for generating a `license-classifications.yml` from ScanCode's license data

fviernau:
Add a Gradle task which generates `license-classifications.yml` from ScanCode's license data hosted at `aboutcode.org` and make use of the generated file.

Fixes #59.

Note: As a follow up, policy rules will need to be extended to handle new license categories.
          For now, the newly added categories will still be flagged by the evaluator rule `UNHANDLED_LICENSE`.

### DIFF
--- a/license-classifications.yml
+++ b/license-classifications.yml
@@ -1,46 +1,71 @@
----
-# Example license-classifications.yml based on categorization from
-# https://github.com/nexB/scancode-toolkit/commit/ed644e4.
 #
-# To demonstrate how one can insert a custom written offer
-# include-source-code-offer-in-notice-file has been set to
-# true for all licenses of the GPL family.
+# License classification generated based on https://scancode-licensedb.aboutcode.org/.
+#
+# This ORT configuration file is provided as an example only. It
+# demonstrates the general configuration capabilities of ORT and does not
+# reflect any real-world configuration used by the ORT contributors, nor
+# are they a recommendation on the configuration to use.
+#
+# Please consult your legal counsel about how ORT should be configured for your use cases.
 #
 # For detailed documentation on this file, see
 # https://github.com/oss-review-toolkit/ort/blob/main/docs/config-file-license-classifications-yml.md.
-
+#
+---
 categories:
+- name: "commercial"
 - name: "copyleft"
 - name: "copyleft-limited"
-- name: "permissive"
-  description: "Licenses with permissive obligations."
-- name: "public-domain"
+- name: "free-restricted"
+- name: "generic"
 - name: "include-in-notice-file"
-  description: >-
-    This category is checked by templates used by the ORT report generator. The licenses associated with this
-    category are included into NOTICE files.
 - name: "include-source-code-offer-in-notice-file"
-  description: >-
-    A marker category that indicates that the licenses assigned to it require that the source code of the packages
-    needs to be provided.
-
+- name: "patent-license"
+- name: "permissive"
+- name: "proprietary-free"
+- name: "public-domain"
+- name: "source-available"
+- name: "unknown"
+- name: "unstated-license"
 categorizations:
-- id: "AGPL-1.0"
+- id: "0BSD"
   categories:
-  - "copyleft"
+  - "permissive"
   - "include-in-notice-file"
-  - "include-source-code-offer-in-notice-file"
+- id: "AAL"
+  categories:
+  - "permissive"
+  - "include-in-notice-file"
+- id: "ADSL"
+  categories:
+  - "permissive"
+  - "include-in-notice-file"
+- id: "AFL-1.1"
+  categories:
+  - "permissive"
+  - "include-in-notice-file"
+- id: "AFL-1.2"
+  categories:
+  - "permissive"
+  - "include-in-notice-file"
+- id: "AFL-2.0"
+  categories:
+  - "permissive"
+  - "include-in-notice-file"
+- id: "AFL-2.1"
+  categories:
+  - "permissive"
+  - "include-in-notice-file"
+- id: "AFL-3.0"
+  categories:
+  - "permissive"
+  - "include-in-notice-file"
 - id: "AGPL-1.0-only"
   categories:
   - "copyleft"
   - "include-in-notice-file"
   - "include-source-code-offer-in-notice-file"
 - id: "AGPL-1.0-or-later"
-  categories:
-  - "copyleft"
-  - "include-in-notice-file"
-  - "include-source-code-offer-in-notice-file"
-- id: "AGPL-3.0"
   categories:
   - "copyleft"
   - "include-in-notice-file"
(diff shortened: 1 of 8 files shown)


license-classifications.yml
```diff
@@ -1,363 +1,7426 @@
...
 - name: "include-source-code-offer-in-notice-file"
-  description: >-
-    A marker category that indicates that the licenses assigned to it require that the source code of the packages
-    needs to be provided.
-
+- name: "patent-license"
+- name: "permissive"
+- name: "proprietary-free"
+- name: "public-domain"
+- name: "source-available"
```
tsteenbe, line 27: Does ScanCode have 2-3 description for each of the categories? Maybe ask @pombredanne. 
Would be nice if we could add a category description as even I struggle to fully understand the source-available and unstated-license categories

sschuberth: BTW, it makes me feel a bit uneasy to solely rely on ScanCode's categorization here. We already have an IMO too tight coupling to ScanCode and its license texts in ORT's SPDX module, and I'd rather not introduce similar couplings elsewhere, but stay scanner-agnostic where we can.

So I'm wondering why we can't do as proposed [here](https://github.com/oss-review-toolkit/ort-config/issues/59#issuecomment-1259281668) and simply use LDBcollector's classification, which at least does not rely on a single source / "opinion" only, and also nicely prefixes (at least some) category names with a hint about *what* the category semantically is about.

fviernau: > BTW, it makes me feel a bit uneasy to solely rely on ScanCode's categorization here. We already have an IMO too tight coupling to ScanCode and its license texts in ORT's SPDX module, and I'd rather not introduce similar couplings elsewhere, but stay scanner-agnostic where we can.

I believe 'ort-config' is about providing a real world example, providing amongst others the following value

1. Users can get started quickly.
2. Users can see how a real world setup could look like.
3. It allows building and sharing a vision of a good config we consider reasonable.
4. Users can see how curations / package configurations and policy rules work together
5. Users can instantly run all stages: analyzer, scanner, evaluator, advisor, reporter and
    see the report with the violations

Regarding point 5: When looking at the report of an arbitrary scan, it will be cluttered with a bunch of UNHANDLED_LICENSE issues. So, it's quite obfuscated how effort / issues is there to fix, as classifying the licenses "maybe" lead to further issues. Having a lot of licenses classified improved that situation quite a bit, the report created by default will be more meaningful. 

In terms of getting the user started quickly, I believe it makes sense to stick to the de-facto default scanner ScanCode for the following reasons:
1. A setup for other scanners would look similar, so it does not illustrate really new cases.
    As it's similar. it's easy for the user to create if the ort-config is understood.
2. Adding categories from mutliple scanners adds complexity. It makes it harder to understand and iterate on the configuration.
3. Categories from mutliple scanners may overlap semantically
4. To have the setup kind of complete, policy rules for all the categories have to be written. 
     Who's gonna do that?

So, at the very minimum, I believe adding categories beyond ScanCode should be considered out of scope of this PR.

tools/curations/buildSrc/src/main/kotlin/ScanCodeLicenseDbClassifications.kt
```diff
@@ -0,0 +1,188 @@
+package org.ossreviewtoolkit.tools.curations
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategies
+import com.fasterxml.jackson.databind.json.JsonMapper
+import com.fasterxml.jackson.module.kotlin.KotlinFeature
+import com.fasterxml.jackson.module.kotlin.KotlinModule
+import com.fasterxml.jackson.module.kotlin.readValue
```
sschuberth, line 7: Given that we're trying to avoid Jackson in ORT Core (also see https://github.com/oss-review-toolkit/ort/issues/3904), would you agree that it also makes sense to rather use kotlinx-serialization here?

tools/curations/buildSrc/src/main/kotlin/ScanCodeLicenseDbClassifications.kt
```diff
@@ -0,0 +1,188 @@
...
+            var license = queue.poll()
+
+            while (license != null) {
+                result[license] = downloadLicenseDetails(license)
+                license = queue.poll()
+            }
+        }
+    }
+
+    val executionTimeMillis = measureTimeMillis {
```
sschuberth, line 116: Better use `measureTime` and let Kotlin automatically choose the unit for the `Duration`'s string representation.

sschuberth (changes requested):

tools/curations/buildSrc/src/main/kotlin/ScanCodeLicenseDbClassifications.kt
```diff
@@ -0,0 +1,188 @@
...
+        threads.run {
+            forEach { it.start() }
+            forEach { it.join() }
+        }
+    }
+
+    LOGGER.quiet("Downloaded ${result.size} files in ${executionTimeMillis / 1000f} seconds on $threadCount threads.")
+
+    return result.toMap()
+}
```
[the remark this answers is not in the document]
fviernau: > Have you considered using the [`download` plugin](https://github.com/michel-kraemer/gradle-download-task) instead to safe some custom code, similar to like [we already do in the `spdx-utils` module](https://github.com/oss-review-toolkit/ort/blob/4aabb3dad4000b985b136471fe9d1cf4b086ee36/utils/spdx/build.gradle.kts#L299-L327)?

I decided against it as it leads to a coupling with Gradle tasks, and seemed in the end a bit more complex.
I found it nicer not to add a dependency for such small thing as parallelizing a download and instead have full control over parallelization.

> In general, Gradle discourages plugins to manually create threads to parallelize work. Instead, the [Worker API](https://docs.gradle.org/current/userguide/worker_api.html) should be used. The download plugin [already does that](https://github.com/michel-kraemer/gradle-download-task/blob/82d1aab0aafffa210ab817ccbcefa18e88464fa4/src/main/java/de/undercouch/gradle/tasks/download/internal/DefaultWorkerExecutorHelper.java).

This task is run manually only very rarely. Given that I believe the manual creation of threads is just a theoretical problem in this case. Would you agree?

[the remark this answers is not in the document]
sschuberth: > This way the code is easily readable even for people who are not into Gradle tasks.

Actually, I found it harder to read (and to check for correctness) than a `Download`-task.

> This task is run manually only very rarely. Given that I believe the manual creation of threads is just a theoretical problem in this case. Would you agree?

Independently of whether the problem is theoretical or real, I believe we should not introduce code that does not adhere to Gradle best practices. But I'm curious about what other @oss-review-toolkit/kotlin-devs say.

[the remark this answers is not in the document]
fviernau: > Actually, I found it harder to read (and to check for correctness) than a Download-task.

I believe that's subjective.

> Independently of whether the problem is theoretical or real, I believe we should not introduce code that does not adhere to Gradle best practices. 

In general I believe if there is a good reason to not follow best practices, that's fine. So, I believe using the fact that something is best practices as rationale why an exception should not be made, does not make much sense.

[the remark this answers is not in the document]
sschuberth: > > Actually, I found it harder to read (and to check for correctness) than a Download-task.
> 
> I believe that's subjective.

Sure it is. Just like your statement before, that the code would be easier to read.

> So, why should Gradle best practice apply to code which is just plain Kotlin without Gradle dependencies?

Because it's actually only being used within Gradle build files, and thus using Gradle infrastructure, which has its own rules. And I currently don't see the code ever being used elsewhere.

> Apart from that, I was arguing that I would like to keep the implementation separate from Gradle.

Why? If you have concrete plans to use the code elsewhere later, please share them so we can make more provident reviews.

MarcelBochtler (commented):
LGTM, but I think that this comment should still be addressed:
https://github.com/oss-review-toolkit/ort-config/pull/64#discussion_r989961319

MarcelBochtler (approved):

PR #65, opened 2022-09-30
Fix a couple of rule violations for the `mime-types` example scan

fviernau:
Fix violations in 4 packages which cause a high amount of false positives, to reduce the amount of errors quite a bit.

### DIFF
--- a/package-configurations/NPM/_/spdx-correct/3.1.0/vcs.yml
+++ b/package-configurations/NPM/_/spdx-correct/3.1.0/vcs.yml
@@ -1,0 +1,20 @@
+---
+id: "NPM::spdx-correct:3.1.0"
+vcs:
+  type: "Git"
+  url: "https://github.com/jslicense/spdx-correct.js.git"
+  revision: "6e556b4af7bee9899e7f16ae8f745719e5517e5d"
+path_excludes:
+- pattern: "package.json"
+  reason: "BUILD_TOOL_OF"
+- pattern: "test.js"
+  reason: "TEST_OF"
+- pattern: "README.md"
+  reason: "DOCUMENTATION_OF"
+license_finding_curations:
+- path: "index.js"
+  reason: "CODE"
+  concluded_license: "Apache-2.0"
+  comment: "The code contains listings of license identifiers which do not apply.\
+    \ The only license text is in the header of the file stating that it is licensed\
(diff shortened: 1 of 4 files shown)


package-configurations/NPM/_/spdx-correct/3.1.0/vcs.yml
```diff
@@ -0,0 +1,20 @@
...
+  reason: "TEST_OF"
+- pattern: "README.md"
+  reason: "DOCUMENTATION_OF"
+license_finding_curations:
+- path: "index.js"
+  reason: "CODE"
+  concluded_license: "Apache-2.0"
+  comment: "The code contains listings of license identifiers which do not apply.\
+    \ The only license text is in the header of the file stating that it is licensed\
+    \ under 'Apache-2.0'."
```
MarcelBochtler, line 20: Should we stick to multi line comments using`|` for consistency?

sschuberth: I tend to agree (not only for consistency, but also as it makes the comment more readable), but to be fair we should probably not insist on it until we have checks in place to enforce this, if we agree on the formatting.

MarcelBochtler: I don't insist on using multi line comments, that's why I just asked the question.

fviernau: Ok, then let's leave the formatting as-is for now. Note: I've formatted it with the `helper-cli` `package config` commands, `format` or `sort` respectively.

MarcelBochtler (approved):

package-configurations/NPM/_/spdx-correct/3.1.0/vcs.yml
sschuberth: > I've formatted it with the `helper-cli` `package config` commands, `format` or `sort` respectively.

So, would you be fine with changing these commands to use `|` as it requires less quoting / escaping?

fviernau: > So, would you be fine with changing these commands to use `|` as it requires less quoting / escaping?

I'd be fine with that.

PR #69, opened 2022-10-21
Use ORT's default name for the rules script

fviernau:
See individual commits.


tsteenbe (approved):

sschuberth:
Just some remarks:
- Committing a symbolic link might breaking the cloning experience for some Windows users (depending on the OS policies).
- I'm fine with exposing the rules from the `evaluator-rules` directory at the root level for direct consumption *if* the rules at `evaluator-rules/src/main/resources/evaluator.rules.kts` actually do something meaningful. I haven't checked on that. So far I saw `evaluator-rules/src/main/resources/evaluator.rules.kts` more like a showcase of what's possible with ORT's evaluator, but what's not necessarily meaningful to do (in a legal / compliance sense).

fviernau:
> * I'm fine with exposing the rules from the `evaluator-rules` directory at the root level for direct consumption _if_ the rules at `evaluator-rules/src/main/resources/evaluator.rules.kts` actually do something meaningful. I haven't checked on that. So far I saw `evaluator-rules/src/main/resources/evaluator.rules.kts` more like a showcase of what's possible with ORT's evaluator, but what's not necessarily meaningful to do (in a legal / compliance sense).

I believe it is meaningful even though not a legal recommendation.

PR #70, opened 2022-10-21
evaluator-rules: Improve the disclaimer a bit

fviernau:
Drop the copy pasted mention of "ORT examples", make the disclaimer a bit more prominent and avoid using the term example overall while still making clear that no legal recommendation is made. While at it wrap it at 80 chars.

### DIFF
--- a/evaluator-rules/src/main/resources/evaluator.rules.kts
+++ b/evaluator-rules/src/main/resources/evaluator.rules.kts
@@ -17,13 +17,11 @@
  * License-Filename: LICENSE
  */
 
-/*******************************************************
- * Example OSS Review Toolkit (ORT) .rules.kts file    *
- *                                                     *
- * Note this file only contains example how to write   *
- * rules. It's recommended you consult your own legal  *
- * when writing your own rules.                        *
- *******************************************************/
+/******************************************************************************
+ * DISCLAIMER: This file only illustrates how to write evaluator rules, it is *
+ *             no recommendation in any way. It is recommended to consult     *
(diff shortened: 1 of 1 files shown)


evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -17,13 +17,11 @@
...
-/*******************************************************
- * Example OSS Review Toolkit (ORT) .rules.kts file    *
- *                                                     *
- * Note this file only contains example how to write   *
- * rules. It's recommended you consult your own legal  *
- * when writing your own rules.                        *
- *******************************************************/
+/******************************************************************************
+ * DISCLAIMER: This file only illustrates how to write evaluator rules, it's  *
+ *             no recommendation in any way. It is recommended to consult     *
```
tsteenbe, line 22: It is recommended -> It's recommended

tsteenbe (changes requested):

evaluator-rules/src/main/resources/evaluator.rules.kts
mnonnenmacher: I think in formal texts "it is" is the better choice than using a contraction.

fviernau: @tsteenbe ok with changing it back again?

fviernau: Changed it anyway as it's minor.

tsteenbe (approved):

PR #73, opened 2022-11-14
Evaluator rules improvements

fviernau:
See individual commits.

Note: Must be merged after https://github.com/oss-review-toolkit/ort/pull/6082.

### DIFF
--- a/evaluator-rules/src/main/resources/evaluator.rules.kts
+++ b/evaluator-rules/src/main/resources/evaluator.rules.kts
@@ -129,6 +129,8 @@ fun getEnabledPolicyRules(): PolicyRules =
         else -> PolicyRules.PROPRIETARY_PROJECT
     }
 
+val enablePriorToOssRules = ortResult.hasLabel("enable-prior-to-oss-rules")
+
 /**
  * Return file path of package configuration in the ORT configuration repository for package [id].
  *
@@ -1365,15 +1367,17 @@ fun RuleSet.commonRules() {
     vulnerabilityWithHighSeverityInPackageRule()
 
     // Prior to open sourcing use case rules (which get executed once):
-    dependencyInProjectSourceRule()
-    missingCiConfigurationRule()
-    missingContributingFileRule()
-    missingGitignoreFileRule()
-    missingLicenseFileRule()
-    missingReadmeFileRule()
-    missingReadmeFileLicenseSectionRule()
-    missingTestsRule()
-    wrongLicenseInLicenseFileRule()
+    if (enablePriorToOssRules) {
+        dependencyInProjectSourceRule()
+        missingCiConfigurationRule()
+        missingContributingFileRule()
+        missingGitignoreFileRule()
+        missingLicenseFileRule()
+        missingReadmeFileRule()
+        missingReadmeFileLicenseSectionRule()
+        missingTestsRule()
+        wrongLicenseInLicenseFileRule()
+    }
 }
 
 fun RuleSet.ossProjectRules() {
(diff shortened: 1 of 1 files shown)


evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -1355,11 +1355,12 @@ fun RuleSet.commonRules() {
     unhandledLicenseRule()
     unmappedDeclaredLicenseRule()
 
+    // ort.yml rules:
```
sschuberth: Nit: Maybe quote like `"ort.yml"`.

evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -1355,11 +1355,12 @@ fun RuleSet.commonRules() {
...
     deprecatedScopeExludeInOrtYmlRule()
-
     packageConfigurationInOrtYmlRule()
     packageCurationInOrtYmlRule()
 
+    // rules for dependencies:
```
sschuberth: Nit: Please start these sentences capitalized.

evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -1240,7 +1240,7 @@ fun RuleSet.packageCurationInOrtYmlRule() = ortResultRule("PACKAGE_CURATION_IN_O
     }
 }
 
-fun RuleSet.vulnerabilityInPackageRule() = packageRule("VULNERABILITY_IN_PACKAGE") {
+fun RuleSet.vulnerabilityInDependencyRule() = packageRule("VULNERABILITY_IN_DEPENDENCY") {
```
sschuberth, line 1245: From a rule-authoring perspective, the new name is a bit misleading as there's also a `dependencyRule`, but you're using a `packageRule`.

evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -129,6 +129,8 @@ fun getEnabledPolicyRules(): PolicyRules =
         else -> PolicyRules.PROPRIETARY_PROJECT
     }
 
+val fileChecksEnabled = ortResult.hasLabel("file-checks-enabled")
```
sschuberth, line 132: IMO better name the label `enable-file-checks`. Or probably yet better, do not name the rule after how the checks are technically implemented, but after the use-case. Like `enable-prior-to-oss-rules` or so.

sschuberth (changes requested):

evaluator-rules/src/main/resources/evaluator.rules.kts
fviernau: Hm, are you saying I should keep the previous name? 
(I found the new one more consistent, as it targets dependencies)
Whether the rule is executed once per edge in the tree or once per dependency IMO is not reflected in any name.

sschuberth: I'm myself unsure what to do here. In rules-context, the term "dependency" actually is a synonym for "package reference". I believe @mnonnenmacher chose "dependencyRule" instead of (the formally more correct) "packageReferenceRule" back then just because it's a simpler term, and we wanted to make writing rules as easy as possible.

So for anyone with that knowledge, it's clear that `dependencyRule`s are executed once per reference to that package in the tree, whereas `packageRule`s are executed once per unique package.

For vulnerabilities, it's enough to check once per unique package. So I'd personally perfer to stick to `vulnerabilityInPackageRule`.

sschuberth: Looks like so, yes. That rule does not seems to make use of any of a dependency's / package reference's properties, i.e. it could (and should, for performance reasons) be a `packageRule` instead of a `dependencyRule`.

fviernau: Ok, I've dropped the commit from this PR. I'll make a follow-up PR for the package rule thing, as I still feel it may be a bit controversial.

sschuberth (approved):

Issue #66, opened 2022-10-10
evaluator-rules: Add handling for new categories in `license-classifications.yml`


fviernau:
The `license-classifications.yml` has been changed to contain ScanCode's license categorizations.
For all new categories, the rules till report UNHANDLED license.

Update the rules so that these new categories are handled appropriately.

PR #74, opened 2022-11-15
evaluator.rules: Apply the copyleft limited rule to any tree level

fviernau:
It does not matter in which tree level the license is found.

Context: #66.


tsteenbe (changes requested):
Currently the rule only check project to its direct dependencies but after this change it will check the whole tree which is computational heavier check.  What if we make a lighter check by using a packageRule with a require `-isProject()`?

fviernau:
> Currently the rule only check project to its direct dependencies but after this change it will check the whole tree which is computational heavier check. What if we make a lighter check by using a packageRule with a require `-isProject()`?

This is currently consistent with the other existing rules, e.g. copyleftInDependencyRule. I'd prefer to address that in a following PR, ok?

fviernau:
> Currently the rule only check project to its direct dependencies but after this change it will check the whole tree which is computational heavier check. What if we make a lighter check by using a packageRule with a require `-isProject()`?

That's impossible as it checks whether it is statically linked.

tsteenbe (approved):

PR #76, opened 2022-11-15
rules: Handle license categories 'proprietary-free' and 'free-restricted'

fviernau:
See individual commits.

Context: #66.

### DIFF
--- a/evaluator-rules/src/main/resources/evaluator.rules.kts
+++ b/evaluator-rules/src/main/resources/evaluator.rules.kts
@@ -41,6 +41,7 @@ fun getLicensesForCategory(category: String): Set<SpdxSingleLicenseExpression> =
 
 val copyleftLicenses = getLicensesForCategory("copyleft")
 val copyleftLimitedLicenses = getLicensesForCategory("copyleft-limited")
+val freeRestrictedLicenses = getLicensesForCategory("free-restricted")
 val permissiveLicenses = getLicensesForCategory("permissive")
 val proprietaryFreeLicenses = getLicensesForCategory("proprietary-free")
 val publicDomainLicenses = getLicensesForCategory("public-domain")
@@ -988,6 +989,13 @@ fun PackageRule.LicenseRule.isException() =
             }
     }
 
+fun PackageRule.LicenseRule.isFreeRestricted() =
+    object : RuleMatcher {
+        override val description = "isFreeRestricted($license)"
+
+        override fun matches() = license in freeRestrictedLicenses
+    }
+
 fun PackageRule.LicenseRule.isProprietaryFree() =
     object : RuleMatcher {
         override val description = "isProprietaryFree($license)"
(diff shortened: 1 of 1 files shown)


evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -1002,8 +1002,9 @@ fun PackageRule.packageManagerSupportsDeclaredLicenses(): RuleMatcher =
  * Policy rules
  */
 
-fun RuleSet.copyleftInDependencyRule() = dependencyRule("COPYLEFT_IN_DEPENDENCY") {
+fun RuleSet.copyleftInDependencyRule() = packageRule("COPYLEFT_IN_DEPENDENCY") {
```
sschuberth, line 1022: IMO, even if a breaking change, the rule name and extension function name should then be adjusted accordingly.

sschuberth: Just to given an example, e.g. here the naming is consistent:

    fun RuleSet.vulnerabilityInPackageRule() = packageRule("VULNERABILITY_IN_PACKAGE")

evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -1242,6 +1251,27 @@ fun RuleSet.packageCurationInOrtYmlRule() = ortResultRule("PACKAGE_CURATION_IN_O
     }
 }
 
+fun RuleSet.proprietaryFreeInDependencyRule() = packageRule("PROPRIETARY_FREE_IN_DEPENDENCY") {
```
sschuberth, line 1282: Also here, neither the function nor the rule name should use the word "dependency".

evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -1144,6 +1152,27 @@ fun RuleSet.deprecatedScopeExludeInOrtYmlRule() = ortResultRule("DEPRECATED_SCOP
     }
 }
 
+fun RuleSet.freeRestrictedInDependencyRule() = packageRule("FREE_RESTRICTED_IN_DEPENDENCY") {
```
sschuberth, line 1155: Also here, neither the function nor the rule name should use the word "dependency".

evaluator-rules/src/main/resources/evaluator.rules.kts
fviernau: To me having the names for these two consistent seems more important, which is why I prefer to
have the name end with `_IN_DEPENDENCY` for both. IMO whether the rule is a package- or dependency rule is an implementation detail which needs not to be reflected in the rule name.

```
fun RuleSet.copyleftInDependencyRule() = packageRule("COPYLEFT_IN_DEPENDENCY") 
fun RuleSet.copyleftLimitedInDependencyRule() = dependencyRule("COPYLEFT_LIMITED_IN_DEPENDENCY") 
```

fviernau: I changed it and ditched my preference.

tsteenbe (approved):

PR #77, opened 2022-11-16
evaluator.rules: Add handling for further license categories

fviernau:
See individual commits.

Context: #66.

### DIFF
--- a/evaluator-rules/src/main/resources/evaluator.rules.kts
+++ b/evaluator-rules/src/main/resources/evaluator.rules.kts
@@ -44,10 +44,27 @@ val contributorLicenseAgreementLicenenses = getLicensesForCategory("contributor-
 val copyleftLicenses = getLicensesForCategory("copyleft")
 val copyleftLimitedLicenses = getLicensesForCategory("copyleft-limited")
 val freeRestrictedLicenses = getLicensesForCategory("free-restricted")
+val genericLicenses = getLicensesForCategory("generic")
 val permissiveLicenses = getLicensesForCategory("permissive")
 val proprietaryFreeLicenses = getLicensesForCategory("proprietary-free")
 val publicDomainLicenses = getLicensesForCategory("public-domain")
 val unstatedLicenses = getLicensesForCategory("unstated-license")
+
+// Set of licenses, which are not acted upon by the below policy rules.
+val ignoredLicenses = listOf(
+    // Contributor License agreement are not relevant for the "use" of open source.
+    "LicenseRef-scancode-generic-cla",
+    // Permissive licenses are not flagged anyway.
+    "LicenseRef-scancode-other-permissive",
+    // Public domain licenses are not flagged anyway.
+    "LicenseRef-scancode-public-domain",
+    // Disclaimers are not relevant because they have no obligations.
+    "LicenseRef-scancode-public-domain-disclaimer",
+    // Public domain licenses are not flagged anyway.
+    "LicenseRef-scancode-us-govt-public-domain",
+    // Disclaimers are not relevant because they have no obligations.
+    "LicenseRef-scancode-warranty-disclaimer"
+).map { SpdxSingleLicenseExpression.parse(it) }.toSet()
 
 /**
  * The complete set of licenses covered by policy rules.
@@ -61,6 +78,7 @@ val handledLicenses = listOf(
     copyleftLicenses,
     copyleftLimitedLicenses,
     freeRestrictedLicenses,
+    genericLicenses,
     permissiveLicenses,
     proprietaryFreeLicenses,
     publicDomainLicenses,
@@ -1005,6 +1023,20 @@ fun PackageRule.LicenseRule.isFreeRestricted() =
         override fun matches() = license in freeRestrictedLicenses
     }
 
+fun PackageRule.LicenseRule.isGeneric() =
+    object : RuleMatcher {
+        override val description = "isGeneric($license)"
+
+        override fun matches() = license in genericLicenses
+    }
+
+fun PackageRule.LicenseRule.isIgnored() =
+    object : RuleMatcher {
+        override val description = "isIgnored($license)"
+
+        override fun matches() = license in ignoredLicenses
+    }
+
 fun PackageRule.LicenseRule.isProprietaryFree() =
     object : RuleMatcher {
         override val description = "isProprietaryFree($license)"
@@ -1201,6 +1233,27 @@ fun RuleSet.freeRestrictedInDependencyRule() = packageRule("FREE_RESTRICTED_IN_D
         error(
             "The dependency ${pkg.metadata.id.toCoordinates()} is licensed under the ScanCode 'free-restricted' " +
                     "categorized license $license. This requires approval.",
+            howToFixLicenseViolationDefault(license.toString(), licenseSource)
+        )
+    }
+}
+
+fun RuleSet.genericInDependencyRule() = packageRule("GENERIC_IN_DEPENDENCY") {
+    require {
+        -isProject()
+        -isExcluded()
+    }
+
+    licenseRule("GENERIC_IN_DEPENDENCY", LicenseView.CONCLUDED_OR_DECLARED_AND_DETECTED) {
+        require {
+            +isGeneric()
+            -isIgnored()
+        }
+
+        error(
+            "The dependency ${pkg.metadata.id.toCoordinates()} might contain a license which is unknown to the " +
+                    " tooling. It was detected as $license which is just a trigger, but not a real license. Please " +
+                    "create a dedicated license identifier if the finding is valid.",
             howToFixLicenseViolationDefault(license.toString(), licenseSource)
         )
     }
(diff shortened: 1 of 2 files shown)


evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -1014,6 +1022,27 @@ fun PackageRule.packageManagerSupportsDeclaredLicenses(): RuleMatcher =
...
+
+    licenseRule("COMMERCIAL_IN_DEPENDENCY", LicenseView.CONCLUDED_OR_DECLARED_AND_DETECTED) {
+        require {
+            +isCommercial()
+            -isExcluded()
+        }
+
+        issue(
+            Severity.ERROR,
+            "The dependency ${pkg.metadata.id.toCoordinates()} is licensed under the ScanCode 'commercial' " +
```
sschuberth, line 1085: I realized too late that we're already doing this in other rules, but isn't it a bit risky to refer to ScanCode as a scanner here? This basically means users should not add commercial licenses from other scanners to the `license-classifications.yml`'s `commercial` category, which creates an unnecessary tight coupling to the contents of `evaluator.rules.kts`.

evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -43,11 +43,22 @@ val commercialLicenses = getLicensesForCategory("commercial")
...
 
+// Set of licenses, which are not acted upon by the below policy rules.
+val ignoredLicenses = listOf(
+    "LicenseRef-scancode-generic-cla",
+    "LicenseRef-scancode-other-permissive",
+    "LicenseRef-scancode-public-domain",
+    "LicenseRef-scancode-public-domain-disclaimer",
+    "LicenseRef-scancode-us-govt-public-domain",
+    "LicenseRef-scancode-warranty-disclaimer"
+).map { SpdxSingleLicenseExpression.parse(it) }.toSet()
```
sschuberth, line 66: Could use `mapTo`.

evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -43,11 +43,22 @@ val commercialLicenses = getLicensesForCategory("commercial")
 val copyleftLicenses = getLicensesForCategory("copyleft")
 val copyleftLimitedLicenses = getLicensesForCategory("copyleft-limited")
 val freeRestrictedLicenses = getLicensesForCategory("free-restricted")
+val genericLicenses = getLicensesForCategory("generic")
```
sschuberth, line 46: On a related note, [ScanCode 31.0.1](https://github.com/nexB/scancode-toolkit/releases/tag/v31.0.1) introduced an `is_generic` flag in its output (see e.g. [here](https://scancode-licensedb.aboutcode.org/generic-cla.json) for `LicenseRef-scancode-generic-cla`). Maybe we should add support for such a flag in `LicenseFinding`, upgrade ScanCode (and the parser to populate that new field), and rely on that property instead of hard-coding license names.

evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -47,16 +47,22 @@ val genericLicenses = getLicensesForCategory("generic")
 val permissiveLicenses = getLicensesForCategory("permissive")
 val proprietaryFreeLicenses = getLicensesForCategory("proprietary-free")
 val publicDomainLicenses = getLicensesForCategory("public-domain")
+val unknownLicenses = getLicensesForCategory("unknown")
```
sschuberth, line 50: Similarly here, we could add support for ScanCode's `is_unknown` to our data model and handle unknown findings generically.

evaluator-rules/src/main/resources/evaluator.rules.kts
fviernau: I believe that for providing a real world example it suffices to use a single scanner, e.g. the de-facto default one in ORT. The license classifications YML is generated and the category 'commercial' refers to ScanCode's commercial category.

evaluator-rules/src/main/resources/evaluator.rules.kts
fviernau: I want to use an immutable set.

evaluator-rules/src/main/resources/evaluator.rules.kts
fviernau: This does rely on the is_generic flag already, see importer code.

evaluator-rules/src/main/resources/evaluator.rules.kts
fviernau: Similar to above.

evaluator-rules/src/main/resources/evaluator.rules.kts
mnonnenmacher: IMO as long as classifications are generated from ScanCode it's good to mention it here as well to make clear where the classification is coming from.

evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -44,11 +44,22 @@ val contributorLicenseAgreementLicenenses = getLicensesForCategory("contributor-
 val copyleftLicenses = getLicensesForCategory("copyleft")
 val copyleftLimitedLicenses = getLicensesForCategory("copyleft-limited")
 val freeRestrictedLicenses = getLicensesForCategory("free-restricted")
+val genericLicenses = getLicensesForCategory("generic")
 val permissiveLicenses = getLicensesForCategory("permissive")
 val proprietaryFreeLicenses = getLicensesForCategory("proprietary-free")
 val publicDomainLicenses = getLicensesForCategory("public-domain")
 val unstatedLicenses = getLicensesForCategory("unstated-license")
 
+// Set of licenses, which are not acted upon by the below policy rules.
```
mnonnenmacher, line 55: I would appreciate if a short version of the explanation why these licenses are ignored from the commit message would also appear here, because this comment is what users will see.

evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -44,11 +44,22 @@ val contributorLicenseAgreementLicenenses = getLicensesForCategory("contributor-
...
+val ignoredLicenses = listOf(
```
mnonnenmacher, line 56: Why is this not the same as `genericLicenses`? After reading the commit message I would have expected that all generic licenses are ignored?

evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -48,16 +48,22 @@ val genericLicenses = getLicensesForCategory("generic")
...
     "LicenseRef-scancode-other-permissive",
     "LicenseRef-scancode-public-domain",
     "LicenseRef-scancode-public-domain-disclaimer",
     "LicenseRef-scancode-us-govt-public-domain",
-    "LicenseRef-scancode-warranty-disclaimer"
+    "LicenseRef-scancode-warranty-disclaimer",
+    // 'unkown' category:
+    "LicenseRef-scancode-free-unknown",
+    "LicenseRef-scancode-license-file-reference",
+    "LicenseRef-scancode-see-license"
```
mnonnenmacher, line 67: Similar to the question in the previous commit, what makes these unknown licenses special so that they shall be ignored but not others from the `unknownLicenses`?

mnonnenmacher (changes requested):

evaluator-rules/src/main/resources/evaluator.rules.kts
fviernau: The set of ignored licenses is currently a real subset of the union of 'generic' and 'unknown'.

It contains only some of the generic licenses which I believe should not be investigated.

evaluator-rules/src/main/resources/evaluator.rules.kts
fviernau: In my view it's waste of time to investigate these. 
Consider e.g. `LicenseRef-scancode-see-license` flags "See license in LICENSE".

How is that actionable? Create a curation which turns the detected license into NONE because the LICENSE file is scanned anyway? ...No, that's waste of time IMO and does add value.

evaluator-rules/src/main/resources/evaluator.rules.kts
fviernau: I've added a comment to each license ID.

evaluator-rules/src/main/resources/evaluator.rules.kts
fviernau: I've removed 'free-unkown' from the list after looking again at it's triggers. 
Even though it flags on things like:  `license: "MIT"`

That gives quite a bit of things to investigate , as the effort is always limited, investigating 'free-unknow' may not be the best place to spend it. Anyhow, let's not ignore it for now.

evaluator-rules/src/main/resources/evaluator.rules.kts
mnonnenmacher: I wasn't obvious why a subset was chosen because the explanation sounds like it applies to generic licenses in general. With the added comments it's clearer.

mnonnenmacher (approved):

tsteenbe (approved):

PR #82, opened 2022-11-17
evaluator.rules: Improve the copyleft in source rules

fviernau:
See individual commits.

Context: #66.

### DIFF
--- a/evaluator-rules/src/main/resources/evaluator.rules.kts
+++ b/evaluator-rules/src/main/resources/evaluator.rules.kts
@@ -1169,16 +1169,9 @@ fun RuleSet.copyleftInSourceRule() = packageRule("COPYLEFT_IN_SOURCE") {
             +isCopyleft()
         }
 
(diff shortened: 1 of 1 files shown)


evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -1071,16 +1071,9 @@ fun RuleSet.copyleftInSourceRule() = packageRule("COPYLEFT_IN_SOURCE") {
             +isCopyleft()
         }
 
-        val message = if (licenseSource == LicenseSource.DETECTED) {
```
tsteenbe, line 1074: Think we should have declared and detected licenses message here.

fviernau: So it produces 
1. The ScanCode 'copyleft' categorized license $license was declared in project $id."
2.  The ScanCode 'copyleft' categorized license $license was detected in project $id."

tsteenbe (approved):

PR #85, opened 2022-11-23
Fix-up logic for isException()

fviernau:


tsteenbe (approved):

evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -968,9 +962,12 @@ fun resolveViolationInSourceCodeText(pkg: Package, license: String) : String {
 }
 
 /**
- * Set of matchers to help keep policy rules easy to understand
+ * Set of matchers to help keep policy rules easy to understand and corresponding helper functions.
  */
 
+fun isException(license: SpdxSingleLicenseExpression): Boolean =
+    ("-exception" in license.toString() && " WITH " !in license.toString())
```
sschuberth, line 969: BTW, couldn't we simply check whether the `SpdxSingleLicenseExpression` is an `SpdxLicenseWithExceptionExpression`?

fviernau: I agree this would make sense to do. Probably in addition to the existing logic. 
On it's own it would not be capably of returning `true` for e.g. `LicenseRef-scancode-ada-linking-exception`.

sschuberth: Actually, according to [this test](https://github.com/oss-review-toolkit/ort/blob/8154fb368be1f438ec1ed953e14c42a6c7e83b2b/utils/spdx/src/test/kotlin/SpdxExpressionTest.kt#L80-L104) I'd expect it to work [if `Strictness` is `ALLOW_LICENSEREF_EXCEPTIONS`](https://github.com/oss-review-toolkit/ort/blob/8154fb368be1f438ec1ed953e14c42a6c7e83b2b/utils/spdx/src/main/kotlin/SpdxExpression.kt#L505-L509).

fviernau: I do not understand how that test relates to that.

sschuberth: The [`dummyExpression` uses e.g. the free text "exception1"](https://github.com/oss-review-toolkit/ort/blob/8154fb368be1f438ec1ed953e14c42a6c7e83b2b/utils/spdx/src/test/kotlin/SpdxExpressionTest.kt#L70) which is [properly recognized as an `SpdxLicenseWithExceptionExpression`](https://github.com/oss-review-toolkit/ort/blob/8154fb368be1f438ec1ed953e14c42a6c7e83b2b/utils/spdx/src/test/kotlin/SpdxExpressionTest.kt#L90C28-L93). So I would expect the same to work for a text like "LicenseRef-scancode-ada-linking-exception".

sschuberth: I've just made that more clear with https://github.com/oss-review-toolkit/ort/pull/6103.

PR #86, opened 2022-11-23
Use the new copyright statement in all files

tsteenbe:
As agreed upon in [1].

[1]: https://github.com/oss-review-toolkit/ort-governance/issues/6


evaluator-rules/build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022 Bosch.IO GmbH
+ * Copyright (C) 2022 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
```
sschuberth, line 2: Shouldn't the URLs here point to the `ort-config` repo? (As you introduce a separate `NOTICE` file in here.)

sschuberth (approved):

PR #87, opened 2022-11-29
notice: Rename directory for notice templates

MarcelBochtler:
By default, macOS uses a case-insensitive file system. Having a `notice` directory and the `NOTICE` file in the same directory makes it hard to work with this repository.


sschuberth:
LGTM, esp. as I would probably run into the same problem on Windows. However, I'm unsure if there's anyone (From EPAM?) relying on the current directory name, so I'd like more people to approve.

fviernau:
I've made the alignments for `orthw`. I believe this PR should wait for thumbs up from @tsteenbe.

mnonnenmacher (approved):

tsteenbe (approved):

PR #88, opened 2022-11-29
ort.yml: Add exclusions for example and tool projects

MarcelBochtler:
These projects are not used for distribution, and can therefore be excluded.

### DIFF
--- a/.ort.yml
+++ b/.ort.yml
@@ -1,0 +1,12 @@
+---
+excludes:
+  paths:
+  - pattern: "tools/**"
+    reason: "BUILD_TOOL_OF"
+    comment: "This directory contains projects used to generate ORT configurations."
+  - pattern: "evaluator-rules/build.gradle.kts"
+    reason: "BUILD_TOOL_OF"
+    comment: "This project is being used to provide IDE support for the evaluator.rules.kts file."
+  - pattern: "notifications/build.gradle.kts"
+    reason: "BUILD_TOOL_OF"
+    comment: "This project is being used to provide IDE support for the example.notifications.kts file."


.ort.yml
```diff
@@ -0,0 +1,12 @@
+---
+excludes:
+  paths:
+  - pattern: "tools/**"
+    reason: "BUILD_TOOL_OF"
+    comment: "This directory contains projects used to generate ORT configurations."
+  - pattern: "evaluator-rules/**"
+    reason: "EXAMPLE_OF"
```
sschuberth, line 8: What's now your official wording here? With all the recent changes from @fviernau to these rules, they seem to go beyond a mere example, into the direction of a recommended default.

Also, if we document this as an example, should we for completeness also declare the `curations` directory and probably others as well as examples, even if not needed from an ORT analysis perspective as these directories do not contain any definition files?

mnonnenmacher: @MarcelBochtler Another option would be to be more specific and just exclude `evaluated-rules/build.gradle.kts` as `BUILD_TOOL_OF`.

fviernau: To me an `ort.yml` usually corresponds to a release artifact. I know we do not have one for this repository, but in the end we should just define what the scan using this ort.yml will be about. My proposal to keep it simple for now:
- assume one single (virtual) release artifact including all ort-config files and only do a single scan. Setup the ort.yml accordingly.

sschuberth: I also believe what @mnonnenmacher and @fviernau propose makes sense: The Gradle projects for rules and notifications are just utilities to get IDE support for both types of scripts when authoring them. The "real thing" are just the scripts themselves. So only the definition files should be excluded.

sschuberth: Gentle ping @MarcelBochtler in case you want to continue working on this now that you're back.

MarcelBochtler: Done. I've now reduced the exclusions for notifications and rules.
However, I kept the exclusion of the tools directory, as this doesn't contain any scan relevant files.

.ort.yml
```diff
@@ -0,0 +1,12 @@
+---
+excludes:
+  paths:
+  - pattern: "tools/**"
+    reason: "BUILD_TOOL_OF"
+    comment: "This directory contains projects used to generate ORT configurations."
+  - pattern: "evaluator-rules/build.gradle.kts"
+    reason: "BUILD_TOOL_OF"
+    comment: "This project is being used to provide IDE support of the rules.kts file."
```
sschuberth, line 9: Nit: Maybe "*for* the evaluator.rules.kts file".

.ort.yml
```diff
@@ -0,0 +1,12 @@
...
+  - pattern: "notifications/build.gradle.kts"
+    reason: "BUILD_TOOL_OF"
+    comment: "This project is being used to provide IDE support of the notifications.kts file."
```
sschuberth, line 12: Nit: Maybe "*for* the example.notifications.kts file".

sschuberth (approved):

PR #93, opened 2023-01-05
Readme: Update copyright year

MarcelBochtler:


fviernau (approved):

README.md
```diff
@@ -43,7 +43,7 @@ This repository is currently in incubation and not yet ready for contributions.
 
 # License
 
-Copyright (C) 2019-2022 [The ORT Project Authors](./NOTICE).
+Copyright (C) 2019-2023 [The ORT Project Authors](./NOTICE).
```
sschuberth, line 46: This reminds me actually of something: Didn't we say that for the generic Copyright statement we do not use year ranges anymore, but only the inception year? At least that's how we do it in the headers in the source code now, as IIRC @tsteenbe mentioned that a Copyright is valid for 100 years from its inception anyway.

MarcelBochtler: I thought that we'd do this only for the ranges in the `NOTICE` file.

PR #94, opened 2023-01-17
curations: Add `source_artifact` for androidx.activity:activity:1.2.0

bs-ondem:


curations/Maven/androidx.activity/activity.yml
```diff
@@ -0,0 +1,9 @@
+- id: "Maven:androidx.activity:activity:1.2.0"
+  curations:
+    comment: |
+      No sources artifact found by ORT due to missing meta data. Sources found via Internet search.
```
sschuberth, line 4: Hmm, a Maven source artifact is never explicitly mentioned in metadata, but the URL to the source artifact is derived from / implied by the URL to the binary artifact. Can you please clarify?

bs-ondem: The analyzer declares the binary artifact for the androidx packages as `RemoteArtifact.EMPTY` and therefore the source artifact is also empty. Because of this, the scanner cannot resolve the provenance for the packages. I thought it would be the right approach to write curations and publish them upstream because there are certainly others who use ORT to scan Android projects.

sschuberth: > The analyzer declares the binary artifact for the androidx packages as `RemoteArtifact.EMPTY`

But isn't this a bug that should be fixed instead? A Maven binary artifact URL should never be empty. PS: I'm looking at similar issues currently in the context of the analysis of a bigger Android project.

PR #98, opened 2023-03-02
fix(notice-templates)!: Re-align with generalization of notice reporter

fviernau:
As of [1] ORT's `PlainTextTemplateReporter`, formerly `NoticeTemplateReporter`, no more adds the `NOTICE_` prefix to the output filename. Instead, the respective filename of the template, without the `.ftl` suffix, is used as output filename. So, add the `NOTICE_` prefix to the template filenames in order to keep that prefix in output filenames.

While at it turn the whole filename to uppercase for consistency. So, `NOTICE_SUMMARY` is now output instead of `NOTICE_summary`.

[1]: https://github.com/oss-review-toolkit/ort/pull/6584


sschuberth (approved):
Not strictly required by this change, but I think it makes sense to also update the ORT version used in this repo to a version that includes the mentioned changes.

fviernau:
> Not strictly required by this change, but I think it makes sense to also update the ORT version used in this repo to a version that includes the mentioned changes.

Addressed by #99.

PR #99, opened 2023-03-02
deps(Gradle): Upgrade ORT to version e50456588d

fviernau:
A preceeding change renamed the notice template files, which was done to align with ORT's recent refactoring of the
`PlainTextTemplateReporter`, formerly called `NoticeTemplateReporter`.

Upgrade ORT to have this `PlainTextTemplateReporter` available.

### DIFF
--- a/tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
+++ b/tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
@@ -44,13 +44,13 @@ open class VerifyPackageCurationsTask : DefaultTask() {
                                 "'${curation.id.toCoordinates()}' in file '$relativePath' does not have a package name."
                     }
 
-                    if (curation.id.version.isNotBlank()) {
+                    val version = curation.id.version
+                    if (version.isNotBlank()) {
                         runCatching {
-                            Requirement.buildIvy(curation.id.version)
+                            Requirement.buildIvy(version)
                         }.onFailure {
-                            issues += "The version '${curation.id.version}' in file '$relativePath' is not a valid " +
-                                    "Ivy version range. See: " +
-                                    "https://ant.apache.org/ivy/history/2.5.0/settings/version-matchers.html"
+                            issues += "The version '$version' in file '$relativePath' is not a valid Ivy version " +
+                            "range. See: https://ant.apache.org/ivy/history/2.5.0/settings/version-matchers.html"
                         }
                     }
 
--- a/tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
+++ b/tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
@@ -48,7 +48,7 @@ open class VerifyPackageCurationsTask : DefaultTask() {
                         runCatching {
                             Requirement.buildIvy(curation.id.version)
                         }.onFailure {
-                            issues += "The version '${curation.id.version}' in file '$relativePath 'is not a valid " +
+                            issues += "The version '${curation.id.version}' in file '$relativePath' is not a valid " +
                                     "Ivy version range. See: " +
                                     "https://ant.apache.org/ivy/history/2.5.0/settings/version-matchers.html"
                         }
--- a/tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
+++ b/tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
@@ -1,13 +1,14 @@
 package org.ossreviewtoolkit.tools.curations
 
 import com.fasterxml.jackson.module.kotlin.readValue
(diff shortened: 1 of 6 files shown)


sschuberth:
Looks like we should also migrate from `com.vdurmont:semver4j` to `org.semver4j:semver4j` here...

tsteenbe (approved):

tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
```diff
@@ -44,12 +44,13 @@ open class VerifyPackageCurationsTask : DefaultTask() {
                                 "'${curation.id.toCoordinates()}' in file '$relativePath' does not have a package name."
                     }
 
-                    if (curation.id.version.isNotBlank()) {
+                    val version = curation.id.version
+                    if (version.isNotBlank()) {
```
sschuberth: Just throwing in some alternatives:
- `with(curation.id)` (even if it adds nesting)
- `curation.id.version.takeUnless { it.isBlank() }?.let { version ->`.

tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
```diff
@@ -44,12 +44,13 @@ open class VerifyPackageCurationsTask : DefaultTask() {
...
                         runCatching {
-                            Requirement.buildIvy(curation.id.version)
+                            Requirement.buildIvy(version)
                         }.onFailure {
-                            issues += "The version '${curation.id.version}' in file '$relativePath' is not a valid " +
-                                    "Ivy version range, see " +
+                            issues += "The version '${version}' in file '$relativePath' is not a valid Ivy version " +
```
sschuberth: `{}` should now be dropped here.

tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
```diff
@@ -48,9 +48,9 @@ open class VerifyPackageCurationsTask : DefaultTask() {
                         runCatching {
                             Requirement.buildIvy(curation.id.version)
                         }.onFailure {
-                            issues += "The version '${curation.id.version}' in file '$relativePath 'is not a valid " +
-                                    "Ivy version range. See: " +
-                                    "https://ant.apache.org/ivy/history/2.5.0/settings/version-matchers.html"
+                            issues += "The version '${curation.id.version}' in file '$relativePath' is not a valid " +
+                                    "Ivy version range, see " +
+                                    "https://ant.apache.org/ivy/history/2.5.0/settings/version-matchers.html."
```
sschuberth: I guess the reason for using the `See:`-style was to *avoid* the need for a dot after the URL, which I know in some environments (like Azure pipelines log output) leads to invalid URLs as the dot it considered a literal part of the URL. Another way to avoid this would be to single-quote the URL. But actually, I'd prefer to stick to the `See:`-style in this case, as it's not part of the fix done in this commit anyway.

tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
```diff
@@ -100,3 +99,6 @@ open class VerifyPackageCurationsTask : DefaultTask() {
         }
     }
 }
+
+private fun String.hasVersionRangeIndicators() =
+    listOf(",", "~", "*", "+", ">", "<", "=", " - ", "^", ".x", "||").any { contains(it, ignoreCase = true) }
```
sschuberth, line 105: Missing final newline.

sschuberth (changes requested):

tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
fviernau: I've kept this one as-is.

tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
```diff
@@ -99,3 +99,6 @@ open class VerifyPackageCurationsTask : DefaultTask() {
...
```
sschuberth, line 105: This recreates the list on each function call, which in this case is on each iteration over `curations`.

Generally, I wonder whether we shouldn't refactor the needed functionality in ORT to be public, and then make the `buildSrc` project depend on the necessary ORT artifact so we can reuse the code from ORT.

sschuberth (dismissed):
Approving despite my comments which could be addressed later.

tools/curations/buildSrc/src/main/kotlin/VerifyPackageCurationsTask.kt
fviernau: > This recreates the list on each function call,

I've fixed it.

sschuberth (approved):

PR #110, opened 2023-04-17
Update license classifications from ScanCode licensedb

### LINKED ISSUE #107: Re-import latest ScanCode license classifications
Meanwhile there have been several upstream changes, so we should re-generate the classifications.

1. A new "cla" category has been introduced which makes our hard-coded category mapping at least partially unnecessary.
2. A bunch of new license have been added
3.Several "unknown" licenses have been removed

fviernau:
See individual commits.

Fixes #107.

### DIFF
--- a/evaluator-rules/src/main/resources/evaluator.rules.kts
+++ b/evaluator-rules/src/main/resources/evaluator.rules.kts
@@ -40,7 +40,7 @@ fun getLicensesForCategory(category: String): Set<SpdxSingleLicenseExpression> =
     }
 
 val commercialLicenses = getLicensesForCategory("commercial")
-val contributorLicenseAgreementLicenenses = getLicensesForCategory("contributor-license-agreement")
+val claLicenses = getLicensesForCategory("cla")
 val copyleftLicenses = getLicensesForCategory("copyleft")
 val copyleftLimitedLicenses = getLicensesForCategory("copyleft-limited")
 val freeRestrictedLicenses = getLicensesForCategory("free-restricted")
@@ -87,7 +87,7 @@ val ignoredLicenses = listOf(
  */
 val handledLicenses = listOf(
     commercialLicenses,
-    contributorLicenseAgreementLicenenses,
+    claLicenses,
     copyleftLicenses,
     copyleftLimitedLicenses,
     freeRestrictedLicenses,
--- a/license-classifications.yml
+++ b/license-classifications.yml
@@ -13,8 +13,8 @@
 #
 ---
 categories:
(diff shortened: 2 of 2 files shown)


fviernau:
@oss-review-toolkit/core-devs:  I just stumbled across the following issue:

1. A particular license ID does not have an SPDX identifier in ScanCode but just a `LicenseRef-*` identifier
2. In a new ScanCode version an Spdx identifier gets added to that license

outcome: ` ./gradlew generateLicenseClassifications` will remove the `LicenseRef-*` identifier in favor of the SPDX one.
This is problematic as one cannot use the `license-classifications.yml` for the older ScanCode version anymore, as the `LicenseRef-*` id would be required there.

Proposal: For each SPDX license add two entries to `license-classifications-yml`: One with SPDX id, one with `LicenseRef-*` id.

evaluator-rules/src/main/resources/evaluator.rules.kts
```diff
@@ -40,7 +40,7 @@ fun getLicensesForCategory(category: String): Set<SpdxSingleLicenseExpression> =
     }
 
 val commercialLicenses = getLicensesForCategory("commercial")
-val contributorLicenseAgreementLicenenses = getLicensesForCategory("contributor-license-agreement")
+val claLicenenses = getLicensesForCategory("cla")
```
sschuberth, line 43: Typo in "Licenenses".

sschuberth (changes requested):

sschuberth:
> Proposal: For each SPDX license add two entries to `license-classifications-yml`: One with SPDX id, one with `LicenseRef-*` id.

How would you actually get both the (old) LicenseRef and the (new) SPDX ID for any specific license if the online ScanCode DB only offers the new data?

fviernau:
> How would you actually get both the (old) LicenseRef and the (new) SPDX ID for any specific license if the online ScanCode DB only offers the new data?

It seems to offer both, see e.g.: https://scancode-licensedb.aboutcode.org/fsf-unlimited-no-warranty.json

sschuberth:
> It seems to offer both, see e.g.: https://scancode-licensedb.aboutcode.org/fsf-unlimited-no-warranty.json

Ok, then the proposed approach makes sense to me.

sschuberth (approved):

fviernau:
> So you would create one entry for the SPDX id and each of the `other_spdx_license_keys`?
> For example Apache-2.0 has more than one: https://scancode-licensedb.aboutcode.org/apache-2.0.json

yes.

PR #114, opened 2023-04-18
package-curations: Scan source artifacts for some `andriodx` and `kotlin` libs

fviernau:
For these artifacts it's better to scan the source artifact, as VCS provenance would need to be figured out manually which is hard in this case.

See individual commits.

### DIFF
--- a/curations/Maven/org.jetbrains.kotlin/kotlin-android-extensions-runtime.yml
+++ b/curations/Maven/org.jetbrains.kotlin/kotlin-android-extensions-runtime.yml
@@ -1,0 +1,5 @@
+- id: "Maven:org.jetbrains.kotlin:kotlin-android-extensions-runtime:"
+  curations:
+    comment: |
+      Enforce scanning the source artifact because figuring out the relevant files in the VCS is complex.
+    source_code_origins: [ ARTIFACT ]
--- a/curations/Maven/androidx.hilt/hilt-common.yml
+++ b/curations/Maven/androidx.hilt/hilt-common.yml
@@ -1,0 +1,6 @@
+- id: "Maven:androidx.hilt:hilt-common:"
+  curations:
+    comment: |
+      Enforce scanning the source artifact, because figuring out the relevant files in the VCS is
+      complex.
+    source_code_origins: [ARTIFACT]
--- a/curations/Maven/androidx.interpolator/interpolator.yml
+++ b/curations/Maven/androidx.interpolator/interpolator.yml
@@ -1,0 +1,6 @@
+- id: "Maven:androidx.interpolator:interpolator:"
+  curations:
+    comment: |
+      Enforce scanning the source artifact, because figuring out the relevant files in the VCS is
+      complex.
+    source_code_origins: [ARTIFACT]
--- a/curations/Maven/androidx.legacy/legacy-support-core-utils.yml
+++ b/curations/Maven/androidx.legacy/legacy-support-core-utils.yml
@@ -1,0 +1,6 @@
+- id: "Maven:androidx.legacy:legacy-support-core-utils:"
+  curations:
+    comment: |
+      Enforce scanning the source artifact, because figuring out the relevant files in the VCS is
+      complex.
+    source_code_origins: [ARTIFACT]
--- a/curations/Maven/androidx.swiperefreshlayout/swiperefreshlayout.yml
+++ b/curations/Maven/androidx.swiperefreshlayout/swiperefreshlayout.yml
@@ -1,0 +1,6 @@
+- id: "Maven:androidx.swiperefreshlayout:swiperefreshlayout:"
+  curations:
+    comment: |
+      Enforce scanning the source artifact, because figuring out the relevant files in the VCS is
+      complex.
+    source_code_origins: [ARTIFACT]
--- a/curations/Maven/org.jetbrains.kotlin/kotlin-annotation-processing-gradle.yml
+++ b/curations/Maven/org.jetbrains.kotlin/kotlin-annotation-processing-gradle.yml
@@ -1,0 +1,5 @@
+- id: "Maven:org.jetbrains.kotlin:kotlin-annotation-processing-gradle:"
+  curations:
+    comment: |
+      Enforce scanning the source artifact because figuring out the relevant files in the VCS is complex.
+    source_code_origins: [ ARTIFACT ]
--- a/curations/Maven/org.jetbrains.kotlin/kotlin-compiler-embeddable.yml
+++ b/curations/Maven/org.jetbrains.kotlin/kotlin-compiler-embeddable.yml
@@ -1,0 +1,5 @@
+- id: "Maven:org.jetbrains.kotlin:kotlin-compiler-embeddable:"
+  curations:
+    comment: |
+      Enforce scanning the source artifact because figuring out the relevant files in the VCS is complex.
+    source_code_origins: [ ARTIFACT ]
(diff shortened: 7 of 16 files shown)


sschuberth:
Have you verified that all of these actually *have* sources artifacts?

fviernau:
> Have you verified that all of these actually _have_ sources artifacts?

Yes, but only for one particular version each.

mnonnenmacher:
I'm not convinced this is a good approach. For one it removes relevant metadata from the packages (or hides it in a curation which is not shown in all reports) which might be important to see outside the scope of scanning. And users can already configure the source origins so that artifacts are preferred over VCS, as in general it should always be more difficult to find relevant files in the VCS than in the source artifacts.

fviernau:
> I'm not convinced this is a good approach. For one it removes relevant metadata from the packages (or hides it in a curation which is not shown in all reports) which might be important to see outside the scope of scanning. 

Do you have something in mind how it could be done nicely? / an alternative solution?

> And users can already configure the source origins so that artifacts are preferred over VCS

AFAIK that configuration option applies to all scans and does not help here IMO.

sschuberth:
Could we maybe go a step back and elaborate what the actual problem is? Is it just a performance issue? Because if these packages have incomplete VCS info, don't we already fall back to scanning the source artifact?

fviernau:
> Could we maybe go a step back and elaborate what the actual problem is? Is it just a performance issue? Because if these packages have incomplete VCS info, don't we already fall back to scanning the source artifact?

1. The Jetbrains ones IIRC do have VCS provenance, but empty VCS path. 
    It's not possible or at least not easy to figure out the right VCS path manually.
    Scanning with empty VCS path is too broad.
2. For the `androidx` ones the tooling detects the following VCS info. Revision cannot be figured out automatically from the 
    Git tags. Figuring it out manually is at minimum painful. 
```
        "type" : "Git",
        "url" : "https://android.googlesource.com/platform/frameworks/support",
        "revision" : "",
        "path" : ""
```
I don't recall it 100% whether the tooling fell back to scanning the source artifact automatically. But highly likely I wouldn't have filed any curation. So, I guess me filing a curation was actually triggered by a Scanner Error failing to download from VCS.

sschuberth:
> There were two different cases (the reason for the curations was not performance):

Thanks for the details.

In these cases, we probably could set the path to a dummy value like (literally) "Override the empty path with a dummy value to trigger an artifact scan instead of scanning the whole repository", would trigger [this code](https://github.com/oss-review-toolkit/ort/blob/59f9eb358ee3923f3d388b12ad40e03a75144280/downloader/src/main/kotlin/VersionControlSystem.kt#L264-L271). That way we explain what we do without overriding any non-empty VCS info.

fviernau:
> In these cases, we probably could set the path to a dummy value like (literally) "Override the empty path with a dummy value to trigger an artifact scan instead of scanning the whole repository", would trigger [this code](https://github.com/oss-review-toolkit/ort/blob/59f9eb358ee3923f3d388b12ad40e03a75144280/downloader/src/main/kotlin/VersionControlSystem.kt#L264-L271). That way we explain what we do without overriding any non-empty VCS info.

1. Are you sure this would then fallback to scanning the source artifact ?
2. If it would indeed fall back, then it would do so only after cloning the VCS, which in these cases indeed is quite a bit of a performance degradation.

sschuberth:
> So, you think the comment in the curation does not make it obvious, because the curation comment is not displayed in the UI right?

I was mainly after not clearing the VCS URL, which might be useful in some cases (like manual investigations).

> Are you sure this would then fallback to scanning the source artifact ?

I'm pretty sure it was like that for the old scanner, but I'm not so sure for the provenance-based scanner. Someone needs to take a look at the `runPathScanners` implementation. Maybe @mnonnenmacher knows out of his head?

mnonnenmacher:
> I'm pretty sure it was like that for the old scanner, but I'm not so sure for the provenance-based scanner. Someone needs to take a look at the `runPathScanners` implementation. Maybe @mnonnenmacher knows out of his head?

That's part of the provenance resolution, if the VCS cannot be resolved it tries to resolve the artifact, depending on this setting:

https://github.com/oss-review-toolkit/ort/blob/8929a30c6dd58f6ec582f3fab66a61134db118dd/model/src/main/kotlin/config/DownloaderConfiguration.kt#L38-L41

> I don't recall it 100% whether the tooling fell back to scanning the source artifact automatically. But highly likely I wouldn't have filed any curation. So, I guess me filing a curation was actually triggered by a Scanner Error failing to download from VCS.

If the fallback to the source artifact does not work this would be a bug that needs to be fixed.

> AFAIK that configuration option applies to all scans and does not help here IMO.
>
> I wanted to tell ORT to scan source artifact only for the given package, and AFAIK there is no other way right now.
We could introduce a more proper way, e.g. allow specifying via curation that only source artifact shall be scanned.
Does this make sense to you?
>
> Until this is done, can we use empty URLs to achieve the same and fix it up later?

If a user wants to do such a configuration that's fine, but IMO we should not provide such workaround curations as part of the ort-config repo.

fviernau:
> For example, if a user only wants to scan VCS but never source artifacts (for whatever reason), this curation would destroy their workflow.

Hm, even in cases where it seems to not make any sense scanning VCS, you want to disallow contributing it to 'ort-config'?
Note: The price is increased onboarding costs, as users will need to figure this out by themselves.

mnonnenmacher:
> Hm, even in cases where it seems to not make any sense scanning VCS, you want to disallow contributing it to 'ort-config'? Note: The price is increased onboarding costs, as users will need to figure this out by themselves.

The point is that package curations are about metadata corrections, not removing metadata. For example, in our case we have prioritized source artifacts over VCS for scanning, so these curations would not bring any value but remove valuable information from the output.

So like I said, this repository should contain metadata corrections that bring value for everyone, not use-case specific workarounds. What's the problem with putting this kind of curations in a private repository?

fviernau:
> So like I said, this repository should contain metadata corrections that bring value for everyone, not use-case specific workarounds. What's the problem with putting this kind of curations in a private repository?

Well, I don't intend to disagree with what you say. It does make sense to me as well. 
The issue I'm trying to solve is the following: If you have user who are willing to contribute but,
then the more use cases you have which cannot be contributed back will increase the effort for contributing back and thus be discouraging users from contributing back. 

I'm just wondering and have a hunch that it may be possible to find a way that even in this use case we find a way to have an entry in ORT config so that the behavior is changed from fitting the (guessed) minority of users to the majority of the users while still providing means for changing behavior explicitly for the minority.

I believe the following would be useful:
1. Extend package curation data by:
```
sourceCodeOriginPriority: List<SourceCodeOrigin>
```
2. Add a curation defining the priority for the package as `listOf(SOURCE_ARTIFACT, VCS)`
3. User who prefer to use `VCS` for the package, can file a curation
    which basically overrides the curation from `ort-config`, e.g. in their PackageOnfigurationProvider with
    higher priority

What do you think?

sschuberth:
> sourceCodeOriginPriority: List<SourceCodeOrigin>

The obvious question I have is: Do we really need to be able to configure this on a per-package level? Or put the other way around: What is the risk in your use-case to configure preference of source artifacts globally?

sschuberth:
As a side note, IMO the right way to deal with these fundamental / platform dependencies for which no source code can be found is to implement https://github.com/oss-review-toolkit/ort/issues/5105.

sschuberth (approved):

curations/Maven/org.jetbrains.kotlin/kotlin-android-extensions-runtime.yml
```diff
@@ -0,0 +1,5 @@
+- id: "Maven:org.jetbrains.kotlin:kotlin-android-extensions-runtime:"
+  curations:
+    comment: |
+      Enforce scanning the source artifact because figuring out the relevant files in the VCS is complex.
+    source_code_origins: [ ARTIFACT ]
```
sschuberth, line 5: Saw this too late, but I guess we should align on not using leading / trailing spaces in list / array definitions?

PR #115, opened 2023-04-18
package-curations: Set a couple of VCS paths

fviernau:
Narrow down the scans of some artifacts build from mono repositories.

### DIFF
--- a/curations/Maven/com.squareup.retrofit2/converter-gson.yml
+++ b/curations/Maven/com.squareup.retrofit2/converter-gson.yml
@@ -1,0 +1,7 @@
+- id: "Maven:com.squareup.retrofit2:converter-gson:"
+  curations:
+    comment: |
+      Set the VCS path of the module inside the multi-module
+      repository.
+    vcs:
+      path: "retrofit-converters/gson"
--- a/curations/Maven/com.squareup.retrofit2/converter-moshi.yml
+++ b/curations/Maven/com.squareup.retrofit2/converter-moshi.yml
@@ -1,0 +1,7 @@
+- id: "Maven:com.squareup.retrofit2:converter-moshi:"
+  curations:
+    comment: |
+      Set the VCS path of the module inside the multi-module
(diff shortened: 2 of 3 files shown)


curations/Pod/_/Firebase%2FCore.yml
```diff
@@ -0,0 +1,7 @@
+- id: "Pod::Firebase/Core:"
```
MarcelBochtler, line 1: Are these curations all for a single repository?
If so, I'd appreciate a link to it in the commit message, to actually check the correctness of the curations.

curations/Pod/_/Firebase%2FCore.yml
```diff
@@ -0,0 +1,7 @@
...
+  curations:
+    comment: |
+      Set the VCS path of the module inside the multi-module
+      repository.
+    vcs:
+      path: "FirebaseCore"
```
MarcelBochtler, line 7: This is not true for all versions.
E.g. a very early version had the Firebase/Core stuff stored in: Firebase/Core
https://github.com/firebase/firebase-ios-sdk/tree/0.0.2/Firebase/Core

curations/Pod/_/GoogleUtilities%2FEnvironment.yml
```diff
@@ -0,0 +1,7 @@
+- id: "Pod::GoogleUtilities/Environment:"
+  curations:
+    comment: |
+      Set the VCS path of the module inside the multi-module
+      repository.
+    vcs:
+      path: "Environment"
```
MarcelBochtler, line 7: Shouldn't this be `GoogleUtilities/Environment`?
https://github.com/google/GoogleUtilities/tree/main/GoogleUtilities/Environment

MarcelBochtler: Same below.

MarcelBochtler (approved):

PR #117, opened 2023-04-18
feat(garbage): Add entries for some Android and iOS projects

fviernau:
These have been produced by ScanCode version 31.2.4.


copyright-garbage.yml
```diff
@@ -1,3 +1,153 @@
...
+- "(c) zbar s:http h:example.com"
+- "(c) ZeI6 VK+-Fs4uO (c) U"
+- "(c) zKA (c) U"
+- "(c) ZTh T"
+- "Apply"
+- "Av z3/4YI&QU5 (c) YEx"
+- "Convert (c), Convert (c)"
+- "Copyright , Example 'let foo"
+- "Copyright (c) $YEAR Mockito contributors"
+- "Copyright (c) $ YEAR Square, Inc."
```
sschuberth, line 120: Hmm. Are esp. these really garbage in the strict sense? It's clearly a Copyright statement, an the company name is apparent, just the year placeholder was not replaced. This certainly is better than nothing if no other statements would be detected, or?

fviernau: For these ones I've verified that the same statement without the palceholder ($) exists. 
However, there is no package context here. If we are very strict package context assumptions would not be allowed.

I've done quite a bit of copyright garbage entries. I usually search the copyright statements for `$` and list them.
For all the cases (plenty) I've verified there was always another copyright statement without the dollar.

Not 100% sure what to do here.

fviernau: > we cannot prove them to be correct by only looking at a single package where it occurred.

Yes. On the other hand, we could say that "templates" are not copyright statements.
It's clearly a trade-off. 
Maybe we should clarify first whether in ort-config we are willing to take small risks here and there. If this is clear discussion will become more efficient I guess.

fviernau: I ditched the controversial entries for now (only to unblock this PR).

copyright-garbage.yml
```diff
@@ -1,3 +1,150 @@
...
+- "OdOA3l (c) Yc"
+- "OkI9 (c) Th1dZEx"
+- "Oo (c) NTh5I"
+- "PSIi0Hu (c) OnUi"
+- "Th (c) Xuu"
+- "U1d (c) BUAa"
+- "Yaml (c) EnumBean"
+- "Yf 2aI OiGAEC/QIaiSS (c) PDDv"
+- "Yt (c) HaeRidI"
+patterns: []
```
sschuberth, line 150: Should we omit this?

sschuberth (approved):

copyright-garbage.yml
fviernau: As this file gets updated programmatically we should also do this: https://github.com/oss-review-toolkit/ort/pull/6879.

PR #123, opened 2023-04-24
VCS path curations for AWS SDK for .NET

MarcelBochtler:


MarcelBochtler:
The version ranges generated by the task are not seen as valid, because Semver4j doesn't support the four Digit version numbers used by the AWS SDK for .NET.

How should I handle this? Make the version range valid by removing the trailing `.0` from the version ranges?

mnonnenmacher:
> How should I handle this? Make the version range valid by removing the trailing `.0` from the version ranges?

This is an option, but you would have to verify that the four digit versions are matched by the curations.

MarcelBochtler:
> > How should I handle this? Make the version range valid by removing the trailing `.0` from the version ranges?
> 
> This is an option, but you would have to verify that the four digit versions are matched by the curations.

Added a test for this: https://github.com/oss-review-toolkit/ort/pull/6922

MarcelBochtler:
> Do we need this code for generating the curations in future?

Yes, for new packages created in this monorepo, this code can be reused.
Additionally, the version ranges might be adjusted if this curation is no longer valid for the open range which is currently used everywhere:
https://github.com/oss-review-toolkit/ort-config/blob/be2fd8918715b852cde26877fb55a6fbac5132dc/tools/curations/buildSrc/src/main/kotlin/Utils.kt#L44-L48

tools/curations/buildSrc/src/main/kotlin/GenerateAwsSdkForNetCurationsTask.kt
```diff
@@ -0,0 +1,44 @@
...
+                    .forEach {
+                        val project = it.substringAfterLast("/")
+                            .removeSuffix(".csproj")
+                            .replace(targetFrameworkRegex, "")
+                        val path = it.substringBeforeLast("/")
+                        val id = Identifier("NuGet::$project")
+
+                        // The AWS SDK for .NET uses a four-part version number for all packages. However, Semver4j
+                        // does not support four-part version numbers. Therefore, remove any trailing zero.
+                        val threeDigitVersionTag = tag.removeSuffix(".0")
```
mnonnenmacher, line 30: Sometimes it's also ".1", so probably better like this: `if( tag.count { it == '.' } == 3) tag.substringBeforeLast(".") else tag`

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #127, opened 2023-05-17
Add curations for "Maven:com.google.code.findbugs:jsr305"

sschuberth:


curations/Maven/com.google.code.findbugs/jsr305.yml
```diff
@@ -0,0 +1,11 @@
+- id: "Maven:com.google.code.findbugs:jsr305:2.0.1"
```
fviernau, line 1: If we followed the approach to scan source artifact this would result in a lot of curations. 
Why not just use a single curation setting the VCS URL to the new github repo, and then
let the tag matching mechanism figure out the provenance automatically?
Well a second one for the VCS path would be needed.

fviernau (changes requested):

curations/Maven/com.google.code.findbugs/jsr305.yml
sschuberth: > Why not just use a single curation setting the VCS URL to the new github repo, and then
> let the tag matching mechanism figure out the provenance automatically?

Because the VCS repo actually does not contain the source code for this release, in no branch, nor tag, nor path.

sschuberth: I stand corrected, they must have published old tags since I've last checked, see https://github.com/findbugsproject/findbugs/tree/2.0.1. I'll change the curation.

sschuberth: Ok, I must have been really confused, as I was already referring to the 2.0.1 tag in the URL... anyway, I've updated the PR now accordingly.

curations/Maven/com.google.code.findbugs/jsr305.yml
```diff
@@ -0,0 +1,19 @@
+- id: "Maven:com.google.code.findbugs:jsr305"
```
fviernau, line 1: Would it makes sense to already now file a VCS path curation ?
e.g. set the path to  "findbugs" ?

curations/Maven/com.google.code.findbugs/jsr305.yml
```diff
@@ -0,0 +1,19 @@
...
+      The original Subversion repository at https://findbugs.googlecode.com/svn/trunk/ does not exist anymore as Google
+      Code was discontinued, so point to the new Git repository at GitHub instead.
+    vcs:
+      type: "Git"
+      url: "https://github.com/findbugsproject/findbugs.git"
+- id: "Maven:com.google.code.findbugs:jsr305:2.0.1"
+  curations:
+    comment: |
+      While no sources artifact has been published for this version, the Git repository contains a "jsr305.jar" file
+      which includes the source code (but also the class files); curate to that source artifact, too, as it is better
```
fviernau, line 13: > curate to that source artifact, too, as it is better than no artifact.

If we put such rationale into the comment, we'll end up with a lot of redundancy. I it should be dropped and the comment further shortened. 

Apart from that, while it may be better for the users, I believe we should also raise the question whether it's worth to maintain such curations. (I'm raising that because in this case scanning the VCS seems to be superior). I'd prefer to drop this curation.

curations/Maven/com.google.code.findbugs/jsr305.yml
```diff
@@ -0,0 +1,19 @@
+- id: "Maven:com.google.code.findbugs:jsr305"
+  curations:
+    comment: |
+      The original Subversion repository at https://findbugs.googlecode.com/svn/trunk/ does not exist anymore as Google
+      Code was discontinued, so point to the new Git repository at GitHub instead.
```
fviernau, line 5: How about shortening the comment, e.g.: "The old subversion repository at ttps://findbugs.googlecode.com/svn/trunk/ has been migrated to Github."

curations/Maven/com.google.code.findbugs/jsr305.yml
fviernau, line 1: commit-msg: Should now say plural 'curations' and not name a version.

fviernau (changes requested):

curations/Maven/com.google.code.findbugs/jsr305.yml
sschuberth: It cannot hurt as a hint, done.

curations/Maven/com.google.code.findbugs/jsr305.yml
sschuberth: > If we put such rationale into the comment, we'll end up with a lot of redundancy.

What kind of redundancy?

I guess what you mean is "duplicate effort", if every curation would get *both* the VCS and artifact curated. However, that's not what I intend to imply or propose here. I simply had put quite some effort into finding an *artifact* for this specific version, and I do not want to loose that knowledge.

curations/Maven/com.google.code.findbugs/jsr305.yml
sschuberth: Done with slightly changed wording.

curations/Maven/com.google.code.findbugs/jsr305.yml
fviernau: The title still mentions the version.

fviernau (changes requested):

curations/Maven/com.google.code.findbugs/jsr305.yml
fviernau: I proposed to drop "curate to that source artifact, too, as it is better..", as that rationale is not specific to that curation but applies to potentially many curations and most importantly does not provide additional relevant information. Such information should go into a contribution guideline IMO instead.

From my experience, people tend to craft curations on existing ones copy and pasting comment. So, I believe it's important to minimize the comments in the first place. (e.g. important for the people maintaining and reviewing it)

fviernau (approved):

PR #129, opened 2023-08-17
feat(evaluator.rules): Flag the use of dependencies without a license

fviernau:


evaluator.rules.kts
```diff
@@ -1354,6 +1354,18 @@ fun RuleSet.missingTestsRule() = projectSourceRule("MISSING_TESTS") {
...
 
+fun RuleSet.noLicenseInDependencyRule() = dependencyRule("NO_LICENSE_IN_DEPENDENCY") {
+    require {
+        -hasLicense()
+        -isExcluded()
+    }
+
+    error(
+        "The dependency '${pkg.metadata.id.toCoordinates()}' does not have any license which is not allowed.",
+        "If the package does have a license, please conclude the license for the package with a curation."
```
sschuberth, line 1365: "does not have any license" sounds a bit odd to me. Maybe rather something like "No license information is available for dependency '${pkg.metadata.id.toCoordinates()}'. If the dependency indeed is unlicensed, it must not be used. Otherwise, please conclude the appropriate license with a package curation."

mnonnenmacher (approved):

PR #130, opened 2023-08-18
feat(curations): Add VCS path curation for kotlin-test-js-runner

MarcelBochtler:
See [1] for the start of the range.

[1]: https://github.com/JetBrains/kotlin/tree/1.4.0/libraries/tools/kotlin-test-js-runner

### DIFF
--- a/curations/Maven/org.jetbrains.kotlin/kotlin-test-js-runner.yml
+++ b/curations/Maven/org.jetbrains.kotlin/kotlin-test-js-runner.yml
@@ -1,0 +1,5 @@
+- id: "Maven:org.jetbrains.kotlin:kotlin-test-js-runner:[1.4.0,)"
+  curations:
+    comment: "Set the VCS path of the module inside the multi-module repository."
+    vcs:
+      path: "libraries/tools/kotlin-test-js-runner"


curations/Maven/org.jetbrains.kotlin/kotlin-test-js-runner.yml
```diff
@@ -0,0 +1,5 @@
+- id: "Maven:org.jetbrains.kotlin:kotlin-test-js-runner:[1.4.0,)"
+  curations:
+    comment: "Set the VCS path of the module inside the multi-module repository."
+    vcs:
+      path: libraries/tools/kotlin-test-js-runner/
```
sschuberth, line 5: While not strictly required, AFAIR we double-quote all paths, so I'd do so here for consistency, too.

curations/Maven/org.jetbrains.kotlin/kotlin-test-js-runner.yml
```diff
@@ -0,0 +1,5 @@
+- id: "Maven:org.jetbrains.kotlin:kotlin-test-js-runner:[1.4.0,)"
+  curations:
+    comment: "Set the VCS path of the module inside the multi-module repository."
+    vcs:
+      path: "libraries/tools/kotlin-test-js-runner/"
```
sschuberth, line 5: Sorry, one more consistency-thing: Existing paths don't end with a "/", so you might want to drop it here, too.

PR #131, opened 2023-08-25
Add and improve curations for Python packages

tsteenbe:
See individual commits for details.


curations/PyPI/_/pycryptodomex.yml
```diff
@@ -0,0 +1,30 @@
...
+    comment: |
+      Mapping declared license based on
+      https://github.com/Legrandin/pycryptodome/blob/v3.9.9/LICENSE.rst,
+      https://github.com/Legrandin/pycryptodome/blob/v3.9.9/setup.py#L480-L482 and
+      https://github.com/Legrandin/pycryptodome/blob/v3.9.9/setup.py#L498.
+    declared_license_mapping:
+      "Apache Software License": "Apache-2.0 AND BSD-2-Clause AND LicenseRef-scancode-public-domain"
+      "BSD License": "Apache-2.0 AND BSD-2-Clause AND LicenseRef-scancode-public-domain"
+      "BSD, Public Domain, Apache": "Apache-2.0 AND BSD-2-Clause AND LicenseRef-scancode-public-domain"
+      "Public Domain": "Apache-2.0 AND BSD-2-Clause AND LicenseRef-scancode-public-domain"
```
MarcelBochtler, line 24: I think these mappings shouldn't include the compound expression. ORT should be able to handle that.

brunopacheco1:
@tsteenbe @MarcelBochtler Hello dear peeps, this is a kind reminder on this PR. We have a couple of projects failing due to PyPI packaging and it would be really great if we can reuse ort-config directly, instead of forking on our side.

curations/PyPI/_/pycryptodomex.yml
sschuberth: @tsteenbe, do you intend to continue working on this PR?

curations/PyPI/_/starlette.yml
```diff
@@ -1,7 +1,9 @@
...
     comment: |
-      The library is licensed under BSD-3-Clause, see:
-      https://github.com/encode/starlette/blob/master/LICENSE.md
+      Mapping declared license based on
+      https://github.com/encode/starlette/blob/0.31.0/LICENSE.md,
+      https://github.com/encode/starlette/blob/0.31.0/pyproject.toml#L10 and
+      https://github.com/encode/starlette/blob/0.31.0/pyproject.toml#L20.
     declared_license_mapping:
-      "BSD License": "BSD-3-Clause"
+      "BSD License": "BSD-2-Clause"
```
sschuberth: I believe this change of the mapping is wrong.

sschuberth (changes requested):
I definitely think there's value in this PR, but we should reorganize the commits for easier review.

Issue #132, opened 2023-08-29
Clarify rules for curations and package configurations


tsteenbe:
Propose to align on set of common rules for curations and package configurations to make it clear and concise for any contributors what is expected and why. Having rules will also help guide the auto generation of these files using orthw.

Below I took a first stab at proposing some rules - I invite all to propose their rules as input for an upcoming ORT TSC discussion.

**General**

Rule: Re-use the same wording for each type of curations and package configuration
Motivation: 
To facilitate automation and ability to understand entries we align on same wording in comments across the ort-config repository each type of curations and package configuration. An example of this would be to use the same sentence for say each VCS path or declared license mapping.

Rule: Use >- over | for YAML multiline strings
Motivation: 
Using a | preserves newlines in YAML but it results in visible \n in the resolved configuration in an ORT analyzer result. Using >- which strips new lines and folds lines does not have this issue

Rule: Links should be immutable
Motivation:
Links in comments are used to provide proof to support the curations or package configurations. If a links is not immutable by pointing to say main, master or develop branch of Git repository which likely changes overtime then verifying the proof becomes difficult.

**Curations**

Rule: Each curation must come with proof
Motivation:
The reason why the curation is needed/valid must be clear to all.

Rule: Individual curations should do only 1 thing
Motivation:
Mixing multiple correction says a VCS url and path curation and a concluded license curations in one block makes review harder and prevents future automation.

Rule: Namespace curations are not allowed
Motivation:
Curations that apply to whole namespaces by only setting the type and namespace of the identifier are not allowed as risk of overwriting valid data due to overbroad package id matching is too big.

Rule: Declared license mapping curations are made without specifying version (range) unless there evidence or relicensing
Motivation:
It’s generally safe to apply a mapping across any version as if a mapping does not map anymore the effect is zero. There is a risk that say a Python library which is ‘BSD licensed’  => BSD-2-Clause becomes invalid when a relicensing to BSD-3-Clause happens. However, from our experience this type of within the same family relicensing is extremely rare. If the package has been relicensed then version ranges should be used to denote the range for which the mapping is valid

Rule: VCS path curations must be for a specific version?
Motivation:
???

**Package Configurations**

Rule: Path excludes should use as broad matching patterns as much as possible
Motivation:
The larger the amount of path excludes the slower ORT’s evaluator/reporter execution gets so use ‘**/tests/**’ instead of specifying tens of absolute directory paths to various tests directories.

[...]

PR #134, opened 2023-09-08
Various minor fixes for the Evaluator rules

tsteenbe:
See individual commits for details.

### DIFF
--- a/evaluator.rules.kts
+++ b/evaluator.rules.kts
@@ -121,7 +121,7 @@ val orgOssProjectsApprovedLicenses = listOf(
 /**
  * Variables used to generate MarkDown text of howToFixDefault()
  */
-var doNotWorryText = "_Note_: Do not worry about creating a perfect curation or exclude. Reviewers will provide feedback."
+val doNotWorryText = "_Note_: Do not worry about creating a perfect curation or exclude. Reviewers will provide feedback."
 val globTutorialMdLink = "[reference documentation](https://docs.oracle.com/javase/tutorial/essential/io/fileOps.html#glob)"
 val ortConfigContributingMdLink = "[CONTRIBUTING.md](https://github.com/oss-review-toolkit/ort-config/blob/main/CONTRIBUTING.md)"
 val ortConfigVcsMdLink = "[ort-config repository](https://github.com/oss-review-toolkit/ort-config)"
--- a/evaluator.rules.kts
+++ b/evaluator.rules.kts
@@ -407,7 +407,7 @@ fun PackageRule.howToFixUnmappedDeclaredLicense(license: String): String {
             |
             |   ```
             |
-            |   curations: Map declared license for ${pkg.metadata.id.toCoordinatesWithoutVersion()}
+            |   feat(curations): Map declared license for ${pkg.metadata.id.toCoordinatesWithoutVersion()}
             |
             |   $relatesToIssueText
             |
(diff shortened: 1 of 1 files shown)


evaluator.rules.kts
```diff
@@ -121,7 +121,7 @@ val orgOssProjectsApprovedLicenses = listOf(
 /**
  * Variables used to generate MarkDown text of howToFixDefault()
  */
-var doNotWorryText = "_Note_: Do not worry about creating a perfect curation or exclude. Reviewers will provide feedback."
+val doNotWorryText = "_Note_: Do not worry about creating a perfect curation or exclude. Reviewers will provide feedback."
```
sschuberth, line 124: Commit message nit: Consider using backticks also for the mention of `doNotWorryText`.

evaluator.rules.kts
```diff
@@ -407,7 +407,7 @@ fun PackageRule.howToFixUnmappedDeclaredLicense(license: String): String {
             |
             |   ```
             |
-            |   curations: Map declared license for ${pkg.metadata.id.toCoordinatesWithoutVersion()}
+            |   feat(curations): Map declared license for ${pkg.metadata.id.toCoordinatesWithoutVersion()}
```
fviernau, line 410: Interresting, wasn't aware that we decided this also for `ort-config`.

fviernau (approved):

evaluator.rules.kts
sschuberth, line 410: While at it, should another commit be added to single-quote the coordinates?

evaluator.rules.kts
```diff
@@ -627,8 +627,8 @@ fun resolveViolationInDependencySourceCodeText(pkg: Package, license: String) :
             |
             |   ```
             |
-            |     For information on how to write a glob pattern, please see this $globTutorialMdLink.
-            |     The available options for the `reason` field are defined in $ortPathExcludeReasonMdLink.
+            |   For information on how to write a glob pattern, please see this $globTutorialMdLink.
```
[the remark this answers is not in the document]
fviernau: I was about to write a similar comment, but then thought the prefix "fix" can be seen as imperative. What do you think about that?

evaluator.rules.kts
sschuberth: Me neither. We're also not enforcing Conventional Commits here. And it also triggers the related question of versioning: Should we start using SemVer here, too? And should each such `feat` commit really bump the minor version? Anyway, I have no objections to do this as proposed for now.

evaluator.rules.kts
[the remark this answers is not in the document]
sschuberth: I was thinking about that, too, but it basically requires you to read the commit type and title as a single sentence, like "Fix (the) incorrect ...", which personally I don't like that much. I prefer to be clear and explicit in the title alone, even if not considering the commit type.

evaluator.rules.kts
fviernau: If this wasn't decided, my tendency would be to not require it already now, because

1. I don't see need for making releases of `ort-config`
2. It adds overhead, e.g. author has to choose, reviewer has to review.
3. Less space for commit message title makes it harder to write a good title

All in all, I believe the added value is slightly negative in sum.

tsteenbe: @fviernau  @sschuberth  See also https://github.com/oss-review-toolkit/ort-config/issues/132#issuecomment-1709595964 - I used conventional commits in the rules as that was used in recent curation commits. Personally I in the middle of using conventional commits for ort-config repository. One the pro it would allow use to make ort-config releases easier just like ort but the negative using it can lead to very long commit messages such as `feat(package-configurations): Add excludes for Maven:net.java.dev.jna:jna-platform:5.8.0`

sschuberth:
@tsteenbe, I guess that merge commit should have been a rebase?

sschuberth (approved):

PR #141, opened 2023-11-14
chore: Change repo config and curation to warning

scottschreckengaust:
A repository configuration or curation within an `.ort.yml` file may be purposeful. This declares a `WARNING` instead of `ERROR` for the evaluator.

Resolves oss-review-toolkit/ort#7753

Please ensure that your pull request adheres to our [contribution guidelines](https://github.com/oss-review-toolkit/.github/blob/main/CONTRIBUTING.md). Thank you!

### DIFF
--- a/evaluator.rules.kts
+++ b/evaluator.rules.kts
@@ -1369,7 +1369,7 @@ fun RuleSet.noLicenseInDependencyRule() = dependencyRule("NO_LICENSE_IN_DEPENDEN
 
 fun RuleSet.packageConfigurationInOrtYmlRule() = ortResultRule("PACKAGE_CONFIGURATION_IN_ORT_YML") {
     if (ortResult.repository.config.packageConfigurations.isNotEmpty()) {
-        error(
+        warning(
             "The use of package configurations is not allowed in the *.ort.yml file.",
             "Please use a global package configuration in the $ortConfigVcsMdLink."
         )
@@ -1378,7 +1378,7 @@ fun RuleSet.packageConfigurationInOrtYmlRule() = ortResultRule("PACKAGE_CONFIGUR
 
 fun RuleSet.packageCurationInOrtYmlRule() = ortResultRule("PACKAGE_CURATION_IN_ORT_YML") {
     if (ortResult.repository.config.curations.packages.isNotEmpty()) {
-        error(
+        warning(
             "The use of package curations is not allowed in the *.ort.yml file.",
             "Please use a global package curation in the $ortConfigVcsMdLink."
         )


evaluator.rules.kts
```diff
@@ -1369,7 +1369,7 @@ fun RuleSet.noLicenseInDependencyRule() = dependencyRule("NO_LICENSE_IN_DEPENDEN
 
 fun RuleSet.packageConfigurationInOrtYmlRule() = ortResultRule("PACKAGE_CONFIGURATION_IN_ORT_YML") {
     if (ortResult.repository.config.packageConfigurations.isNotEmpty()) {
-        error(
+        warning(
             "The use of package configurations is not allowed in the *.ort.yml file.",
```
sschuberth, line 1373: Actually, I'd propose to add a second commit to this PR which also changes the wording to say

> The use of package configurations in the *.ort.yml file is currently disabled. Either enable the feature in the ORT configuration, or move package configurations to the $ortConfigVcsMdLink.

The current wording is overly strict IMO, and makes it sound as if the use of project-local package configurations would be discouraged or so. The same applies to the wording for package curations below.

sschuberth:
Commit message:

> Closes #7753

This needs a [different syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) as it's supposed to close an issue in a different repo: `Closes oss-review-toolkit/ort#7753.`

Also, the commit message body needs to be hard-wrapped at column 75.

evaluator.rules.kts
scottschreckengaust: How about:
> Package configurations found in the *.ort.yml file. The recommended location for package configurations is $ortConfigVcsMdLink.

There is no way (that I know of) enable the feature in the ORT configuration.

sschuberth: > There is no way (that I know of) enable the feature in the ORT configuration.

There [is](https://github.com/oss-review-toolkit/ort/blob/bb742aa3cd8c47fd33ca47a282b883b09522283f/model/src/main/resources/reference.yml#L31-L32).

fviernau: > The current wording is overly strict IMO, and makes it sound as if the use of project-local package configurations would be discouraged or so. The same applies to the wording for package curations below.

My intention when writing this rule was exactly that: A rule which discourages the use of package configurations in `ort.yml`. I believe someone who does not want to discourage that, should simply not use this rule.

scottschreckengaust: > > There is no way (that I know of) enable the feature in the ORT configuration.
> 
> There [is](https://github.com/oss-review-toolkit/ort/blob/bb742aa3cd8c47fd33ca47a282b883b09522283f/model/src/main/resources/reference.yml#L31-L32).

I believe both properties, `enableRepositoryPackageCurations` and `enableRepositoryPackageConfigurations` are set to "true" with the below GitHub action, yet result is the same here:
https://github.com/scottschreckengaust/orter/actions/runs/6870755788/job/18686264762#step:3:1445

```
---
name: ort-toolkit
on:
  push:
    branches:
      - main
  workflow_dispatch: {}
jobs:
  ort:
    runs-on: ubuntu-latest
    permissions:
      contents: write
    steps:
      - name: Checkout project
        uses: actions/checkout@v3
      - name: Run GitHub Action for ORT
        uses: oss-review-toolkit/ort-ci-github-action@7f23c1f8d169dad430e41df223d3b8409c7a156e
        with:
          ort-cli-args: '-P ort.forceOverwrite=true -P ort.enableRepositoryPackageConfigurations=true -P ort.enableRepositoryPackageCurations=true --stacktrace'
```

scottschreckengaust: I added comment as requested. The ability to enable repository configurations and curations is a separate issue.

sschuberth: > My intention when writing this rule was exactly that: A rule which discourages the use of package configurations in `ort.yml`. Not wanting to discourage that is a valid case for sure, but in that case a user should simply not use this rule.

But isn't the goal of the rules in this repo to stay as much as possible unopinionated WRT personal preferences / company policies, and focus on undisputable cases to give examples?

fviernau: > But isn't the goal of the rules in this repo to stay as much as possible unopinionated WRT personal preferences / company policies, and focus on undisputable cases to give examples?

I believe that it can be opinionated with regard to best practices. This thing I would consider a best practice thing.

sschuberth: > This thing I would consider a best practice thing.

See, I wouldn't. We've added the feature for a purpose, and it's a valid thing to do in some workflows (like GitHub actions). While we should point out the possible implications, and also warn if there are local curations / configurations while the feature is disabled globally, I do not consider it to be a bad practice to make use of that feature.

> If it fits most of the cases, why not put it in here?

It fitting "most of the cases" is an assessment that I do not share.

fviernau: I'm fine to disagree.

sschuberth: You mean without also changing the wording of the messages? Yes, that would work as a compromise, until we're more clear about the practices to follow for this repo.

fviernau: > You mean without also changing the wording of the messages? 

Yes.

sschuberth: So, sorry @scottschreckengaust, would you then mind dropping your second commit again? Any while at it, probably change the first commit's title to use `chore` (or `refactor`) as a type rather than `feat`. Thanks in advance!

scottschreckengaust: removed re-wording commit, and switched from `feat` -> `chore` in commit

sschuberth (approved):

PR #145, opened 2023-11-17
chore(gradle): Upgrade Gradle wrapper to version 8.4

fviernau:


sschuberth:
The `distributionSha256Sum` needs to be updated as well. But I'd propose to wait for Gradle 8.5 which should be released very soon.

fviernau:
> The `distributionSha256Sum` needs to be updated as well. But I'd propose to wait for Gradle 8.5 which should be released very soon.

Can we merge this, so that it does not need to hang around waiting?

sschuberth (approved):

PR #147, opened 2023-11-23
curations: Map declared license for PyPI::pandas and PyPI::numpy

inderps:
Screenshot 2023-11-23 at 17 04 28
Screenshot 2023-11-23 at 17 04 44

### DIFF
--- a/curations/PyPI/_/numpy.yml
+++ b/curations/PyPI/_/numpy.yml
@@ -1,0 +1,7 @@
+- id: "PyPI::numpy"
+  curations:
+    comment: |
+      The library is licensed under BSD-3-Clause, see:
+      https://github.com/numpy/numpy/blob/main/LICENSE.txt
+    declared_license_mapping:
+      "BSD License": "BSD-3-Clause"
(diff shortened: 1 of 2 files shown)


curations/PyPI/_/numpy.yml
```diff
@@ -0,0 +1,7 @@
+- id: "PyPI::numpy"
+  curations:
+    comment: |
+      The library is licensed under BSD-3-Clause, see:
+      https://github.com/numpy/numpy/blob/main/LICENSE.txt
```
sschuberth, line 5: Please use permalinks here and below to avoid contents of the link to change. Like in this case, probably the first commit here the file was introduced: https://github.com/numpy/numpy/blob/67e756592f6ebec479e404a104380ca373ad11f1/LICENSE.txt

inderps: Closing this PR because i was informed by Thomas Steenbergen (tsteenbe) that there is already a pending PR which will add all these curations in bulk

PR #150, opened 2023-12-11
chore: Add PyPI::mock curation to version 5.1.0

brunopacheco1:
The package's metadata data only declares an unspecific "BSD License",
but a closer look at the LICENSE.txt reveals that concretely always
the "BSD-2-Clause" license has been used.

### DIFF
--- a/curations/PyPI/_/mock.yml
+++ b/curations/PyPI/_/mock.yml
@@ -1,0 +1,9 @@
+- id: "PyPI::mock:(,5.1.0["
+  curations:
+    comment: |
+      The package's metadata data only declares an unspecific "BSD License", 
+      but a closer look at the LICENSE.txt reveals that concretely always
+      the "BSD-2-Clause" license has been used.
+      https://github.com/testing-cabal/mock/blob/d344fa2794b3b1ae7e4a4dbf265fb040d6f41d1f/LICENSE.txt
+    declared_license_mapping:
+      "BSD License": "BSD-2-Clause"


sschuberth:
@brunopacheco1 please have a look at the failed verification check: In order to not fall into the trap of giving legal advice, curations in this repository must not directly contain concluded licenses. Generally, explicitly concluding licenses via curations is a "sledge hammer" method as it causes all declared and detected license information to be ignored, and it is completely nontransparent how the conclusion was achieved.

Thus, we unfortunately cannot accept the contribution as-is.

brunopacheco1:
@sschuberth True, it was a mistake on my side. The PR is fixed.

curations/PyPI/_/mock.yml
```diff
@@ -0,0 +1,7 @@
+- id: "PyPI::mock:(,5.1.0["
+  curations:
+    comment: |
+      The library is licensed under BSD-2-Clause, see:
+      https://github.com/testing-cabal/mock/blob/d344fa2794b3b1ae7e4a4dbf265fb040d6f41d1f/LICENSE.txt
+    declared_license_mapping:
+      "BSD License": "BSD-2-Clause"
```
sschuberth, line 7: Your curation IMO does not really match what you're saying in the commit message:

> OSS Review Toolkit did not find a license for `PyPI::mock`.

If the ORT analyzer wouldn't have been able to find [the declared license](https://github.com/testing-cabal/mock/blob/d344fa2794b3b1ae7e4a4dbf265fb040d6f41d1f/setup.cfg#L12) there would be no point in mapping it.

So actually, I believe what you want to say is:

> The package's metadata data only declares an unspecific "BSD License", but a closer look at the `LICENSE.txt` reveals that concretely always the "BSD-2-Clause" license has been used.

sschuberth (changes requested):

curations/PyPI/_/mock.yml
brunopacheco1: Indeed your suggestion is more appropriate, I will update comment and commit message.

 Regarding the version, is it advisable to add version always? Or can we have exceptions? This case the license has not changed yet.

sschuberth: Regarding the version, it's advisable to at least add an upper bound of the current version, since we cannot foresee if the future the license would change.

brunopacheco1: Changes done accordingly.

sschuberth (approved):

PR #151, opened 2023-12-11
chore: Add PyPI::python-dateutil curation to version 2.8.2

brunopacheco1:
The license of python-dateutil has changed, as it can be seeing
in the curations comment. We are mapping BSD License to to
BSD-3-Clause, and Dual License to `Apache-2.0 AND BSD-3-Clause`

### DIFF
--- a/curations/PyPI/_/python-dateutil.yml
+++ b/curations/PyPI/_/python-dateutil.yml
@@ -1,0 +1,16 @@
+- id: "PyPI::python-dateutil:(,2.8.2["
+  curations:
+    comment: |
+      The LICENSE file states that contributions after 2017-12-01 as well as older
+      contributions that have been re-licensed fall under Apache-2.0, and non-relicensed
+      older contributions fall under BSD-3-Clause. However, it also
+      says that "The above BSD License Applies to all code, even that also covered by Apache 2.0" [1],
+      meaning that all code in total falls under Apache-2.0 AND BSD-3-Clause.
+      This seems to contradict the declared "Dual License" [2], which usually implies a choice between
+      the licenses. So even if Apache-2.0 OR BSD-3-Clause might have been intended, the wording
+      says something else, and sticking to Apache-2.0 AND BSD-3-Clause is the conservative conclusion.
+      [1]: https://github.com/dateutil/dateutil/blob/0586f4afa26fc6799128d98d4f97a49c7d6ab314/LICENSE#L54C1-L54C81
+      [2]: https://github.com/dateutil/dateutil/blob/0586f4afa26fc6799128d98d4f97a49c7d6ab314/setup.cfg#L16
+    declared_license_mapping:
+      "BSD License": "BSD-3-Clause"
+      "Dual License": "Apache-2.0 AND BSD-3-Clause"


curations/PyPI/_/python-dateutil.yml
```diff
@@ -0,0 +1,12 @@
+- id: "PyPI::python-dateutil:"
+  curations:
+    comment: |
+      Current license: https://github.com/dateutil/dateutil/blob/master/LICENSE
```
sschuberth, line 4: Whenever referring to GitHub content please use permalinks, as otherwise the page content might change over time.

Ideally, there would be one permalink to license content before December 1, 2017, and another permalink for after that date, clearly showing what "BSD License" and "Dual License" refer to.

sschuberth (changes requested):

curations/PyPI/_/python-dateutil.yml
brunopacheco1: @sschuberth I have updated the comment accordingly.

curations/PyPI/_/python-dateutil.yml
```diff
@@ -0,0 +1,13 @@
+- id: "PyPI::python-dateutil:(,2.8.2["
+  curations:
+    comment: |
+      All contributions after December 1, 2017 released under dual license
+      - either Apache 2.0 License or the BSD 3-Clause License.
```
sschuberth, line 5: Hmm, I'm actually reading the new `LICENSE` file differently. WRT the added Apache-2.0 license it says:

> The above license applies to all contributions after 2017-12-01, as well as
all contributions that have been re-licensed

which to me means that there are portions in the existing code base that were *not* re-licensed to Apache-2.0. Also, below the old BSD-3 license it says:

> The above BSD License Applies to all code, even that also covered by Apache 2.0.

So to me, this is not an `OR`-choice that we have, but and `AND` condition.

@willebra, I'd be interested in your opinion here.

willebra: I looked at the package, the key files in the root folder, readme, authors, license, contributing etc. I did not do any scans but looked at some central src files too. The central src files I looket at did not contain any meaningful licensing info, so the files in root folder are the remaining (ok) source for the license info. I also looked at the commit from 6 years back, which was just a plain BSD-3-Clause license.

While the actual situation seems a bit more complicated, I would curate this package, as a whole, as "Apache-2.0 AND BSD-3-Clause".

Some comments, in case someone at the project looks at this. The project's communication is not super-exact, but they likely want to currently use either "Apache-2.0" or "Apache-2.0 OR BSD-3-Clause" license statements for as much of the code as possible. But since there are earlier contributions that are solely under BSD-3-Clause and which have not been relicensed to the new licensing scheme, a conservative yet workable approach would be to treat the whole package as "Apache-2.0 AND BSD-3-Clause". Individual files could be "BSD-3-Clause", "Apache-2.0" or "Apache-2.0 OR BSD-3-Clause". Some files could also be "Apache-2.0 AND BSD-3-Clause" due to relicensing of old files by the project under "Apache-2.0". All in all, calculating this combination leads to my suggestion above. Since it doesn't seem a problematic outcome, and is conservative, I wouldn't spend more time on this.

brunopacheco1: As suggested I changed the `Dual License` to `Apache-2.0 AND BSD-3-Clause`.

sschuberth: Ok, but now also the curation's comments needs some updating. I propose something along the lines of this:

> The LICENSE file states that contributions after 2017-12-01 as well as older contributions that have been re-licensed fall under `Apache-2.0`, and non-relicensed older contributions fall under `BSD-3-Clause`. However, it also says that "The above BSD License Applies to all code, even that also covered by Apache 2.0" [1], meaning that all code in total falls under `Apache-2.0 AND BSD-3-Clause`. This seems to contradict the declared "Dual License" [2], which usually implies a choice between the licenses. So even if `Apache-2.0 OR BSD-3-Clause` might have been intended, the wording says something else, and sticking to `Apache-2.0 AND BSD-3-Clause` is the conservative conclusion.
> [1]: https://github.com/dateutil/dateutil/blob/0586f4afa26fc6799128d98d4f97a49c7d6ab314/LICENSE#L54C1-L54C81
> [2]: https://github.com/dateutil/dateutil/blob/0586f4afa26fc6799128d98d4f97a49c7d6ab314/setup.cfg#L16

curations/PyPI/_/python-dateutil.yml
```diff
@@ -0,0 +1,13 @@
...
+      License after December 1, 2017: https://github.com/dateutil/dateutil/blob/0586f4afa26fc6799128d98d4f97a49c7d6ab314/LICENSE
+      
+      Contributions before December 1, 2017 - except those those explicitly relicensed 
+      - are released only under the BSD 3-Clause License.
+      License before December 1, 2017: https://github.com/dateutil/dateutil/blob/1c8c90338a696c8c146c7eb6cc1bf52016912852/LICENSE
+    declared_license_mapping:
+      "BSD License": "BSD-3-Clause"
```
sschuberth, line 12: Just wondering, why isn't also a mapping for the [declared](https://github.com/dateutil/dateutil/blob/0586f4afa26fc6799128d98d4f97a49c7d6ab314/setup.cfg#L22C32-L22C55) "Apache Software License" required here?

curations/PyPI/_/python-dateutil.yml
sschuberth: Also the commit message needs some work. If we have such an elaborate comment, IMO the commit message body can be left empty in exchange, and the commit message title could just say

    chore: Add license mappings for python-dateutil to up 2.8.2

sschuberth (changes requested):

curations/PyPI/_/python-dateutil.yml
brunopacheco1: If you check the results from this build, there is no mention to it.

 https://github.com/GenomicDataInfrastructure/gdi-userportal-ort/actions/runs/7153573748

image

sschuberth: Yeah, but *why*? Can the full analyzer result be downloaded from somewhere?

curations/PyPI/_/python-dateutil.yml
```diff
@@ -0,0 +1,16 @@
...
+  curations:
+    comment: |
+      The LICENSE file states that contributions after 2017-12-01 as well as older
+      contributions that have been re-licensed fall under Apache-2.0, and non-relicensed
+       older contributions fall under BSD-3-Clause. However, it also 
+       says that "The above BSD License Applies to all code, even that also covered by Apache 2.0" [1], 
+       meaning that all code in total falls under Apache-2.0 AND BSD-3-Clause.
+       This seems to contradict the declared "Dual License" [2], which usually implies a choice between 
+       the licenses. So even if Apache-2.0 OR BSD-3-Clause might have been intended, the wording 
+       says something else, and sticking to Apache-2.0 AND BSD-3-Clause is the conservative conclusion.
```
sschuberth, line 11: Is the indentation different here on purpose?

curations/PyPI/_/python-dateutil.yml
fviernau: > Just wondering, why isn't also a mapping for the [declared](https://github.com/dateutil/dateutil/blob/0586f4afa26fc6799128d98d4f97a49c7d6ab314/setup.cfg#L22C32-L22C55) "Apache Software License" required here?

This is probably because the mapping is built into ORT.

brunopacheco1: In the link I shared before you could get the output of ORT, but I upload the file here.

[analyzer-result.json](https://github.com/oss-review-toolkit/ort-config/files/13661854/analyzer-result.json)

curations/PyPI/_/python-dateutil.yml
brunopacheco1: Good catch, I don't know what happen but it's fixed now.

curations/PyPI/_/python-dateutil.yml
sschuberth: > This is probably because the mapping is built into ORT.

Oh, [indeed](https://github.com/oss-review-toolkit/ort/blob/7aa48950ca1f2fa899a622288a750a4be84c7b64/utils/spdx/src/main/resources/declared-license-mapping.yml#L123C10-L123C10). I had checked before, but only found [the plural form](https://github.com/oss-review-toolkit/ort/blob/7aa48950ca1f2fa899a622288a750a4be84c7b64/utils/spdx/src/main/resources/declared-license-mapping.yml#L25) and did not look further in the "Non-ambiguous mappings" section, as it actually *is* ambiguous (just like the plural form is).

sschuberth (approved):

PR #153, opened 2023-12-14
chore: Add source artifact curations for ANTLR 2.7.6 and 2.7.7

sschuberth:


curations/Maven/antlr/antlr.yml
```diff
@@ -0,0 +1,18 @@
...
+      Neither SCM metadata nor a published source artifact is available. The artifact was found via Internet search.
+    source_artifact:
+      url: "http://www.antlr2.org/download/antlr-2.7.6.tar.gz"
+      hash:
+        value: "e923acd81c590234ea241da96a536e3dcf7f77fb"
+        algorithm: "SHA1"
+- id: "Maven:antlr:antlr:2.7.7"
+  curations:
+    comment: |
+      Neither SCM metadata nor a published source artifact is available. The artifact was found via Internet search.
```
fviernau, line 13: "published source artifact is available": Isn't the source artifact published, just not in the MVN repo? 

Maybe this can be re-phrased ?

sschuberth: > Maybe this can be re-phrased?

Done.

> Shoud we consider mentioning SVM metadata unrelated?

Again, I'd like to emphasize that this source artifact is our only option to get scan results at all, even for people that usually prefer VCS sources code origins over sources artifacts.

fviernau: > Again, I'd like to emphasize that this source artifact is our only option to get scan results at all, even for people that usually prefer VCS sources code origins over sources artifacts.

Would you agree that it is always ok to file a curation which sets

1. Missing source artifact URL
2. Missing VCS info

...independently of the actual value of the "other" provenance info?

If we decide the answer for this question should be "yes", then I believe the sentence in question should in fact be dropped.

sschuberth: I agree that setting either is better than not having that curation at all, and I'd personally accept such a curation even if it might be "incomplete", as it at least improves the situation. However, personally I'd always strive for making the curation "complete" as it usually is not much more effort to determine a source code origin if you already have found he other one.

fviernau (approved):

PR #154, opened 2023-12-14
chore: Add a VCS / path curation for `flexmark-util`

sschuberth:


curations/Maven/com.vladsch.flexmark/flexmark-util.yml
```diff
@@ -0,0 +1,7 @@
+- id: "Maven:com.vladsch.flexmark:flexmark-util"
+  curations:
+    comment: |
+      No sources artifact published and no <scm> tag in POM. Repository found via Internet search.
```
fviernau, line 4: Would it make sense to drop "No sources artifact published" as irrelevant? Maybe also "Repository found via Internet search." ?

idea: we could align on "The package metadata lacks the VCS info." ?

fviernau (changes requested):

curations/Maven/com.vladsch.flexmark/flexmark-util.yml
sschuberth: I've reused the wording from the existing https://github.com/oss-review-toolkit/ort-config/blob/main/curations/Maven/com.vladsch.flexmark/flexmark-all.yml which I wanted to align with.

I believe it's not irrelevant that no sources artifact is published, otherwise the question could arise why it's not being used. This makes more clear that fixing the VCS info is our only option, even for users that have configured ORT to prefer source artifacts over VCS locations.

So, whatever wording change we eventually do in the end, it should be applied to `flexmark-all`, too.

sschuberth: > Would it make sense to drop "No sources artifact published" as irrelevant? Maybe also "Repository found via Internet search." ?

After our discussion, I've dropped the first sentence but kept the second, as to me it's important to explain *how* a conclusion / curation was achieved. That would be unclear when just stating "The package metadata lacks the VCS info".

fviernau (approved):

PR #158, opened 2023-12-19
chore: Fix version matcher of some pypi curation files

brunopacheco1:


brunopacheco1:
@sschuberth I opened recently some PR's, but it turned out I did not pay close attention to the version matcher. It should be `lower or equals to` instead.

sschuberth:
> It should be `lower or equals to` instead.

Makes sense to include the latest releases (for which we can verify the license) in the range. Thanks for the fixup!

PR #159, opened 2023-12-19
chore: Map a declared license for PyPI::pytest-cov

brunopacheco1:
The library is licensed under MIT, but according to
      this [Pull Request](https://github.com/pytest-dev/pytest-cov/pull/467),
      the project metadata was mistakenly stating `BSD License`.
      All versions below 2.12.0 are impacted by this issue.
      Current (4.1.0) license: https://github.com/pytest-dev/pytest-cov/blob/2c9f2170d8575b21bafb6402eb30ca7de31e20b9/LICENSE
      Current project metadata: https://github.com/pytest-dev/pytest-cov/blob/2c9f2170d8575b21bafb6402eb30ca7de31e20b9/setup.py
      License at 2.12.0: https://github.com/pytest-dev/pytest-cov/blob/9692bad8c3501b77cf950d0732ae9cb5c8bb0bd4/LICENSE
      Project metadata at 2.12.0: https://github.com/pytest-dev/pytest-cov/blob/9692bad8c3501b77cf950d0732ae9cb5c8bb0bd4/setup.py
      License before 2.12.0: https://github.com/pytest-dev/pytest-cov/blob/5e1913e013eb06d5bd1357695349d0a75fbf0503/LICENSE
      Project metadata before 2.12.0: https://github.com/pytest-dev/pytest-cov/blob/5e1913e013eb06d5bd1357695349d0a75fbf0503/setup.py

### DIFF
--- a/curations/PyPI/_/pytest-cov.yml
+++ b/curations/PyPI/_/pytest-cov.yml
@@ -1,0 +1,8 @@
+- id: "PyPI::pytest-cov:(,2.12.0["
+  curations:
+    comment: |
+      The LICENSE file always contained only the 'MIT' license text, while the project metadata accidentally
+      declared it to be licensed under 'BSD License' for all version prior to 2.12.0, see also
+      https://github.com/pytest-dev/pytest-cov/pull/467.
+    declared_license_mapping:
+      "BSD License": "MIT"


sschuberth:
> The library is licensed under MIT, but according to
> this [Pull Request](https://github.com/pytest-dev/pytest-cov/pull/467),
> the project metadata was mistakenly stating `BSD License`.

That's an interesting case. What's your take on this @willebra? Does this really make past releases MIT- instead of BSD-licensed?

PS: I've [asked](https://github.com/pytest-dev/pytest-cov/pull/467#issuecomment-1862764363) the maintainers to be explicit on that.

curations/PyPI/_/pytest-cov.yml
```diff
@@ -0,0 +1,15 @@
+- id: "PyPI::pytest-cov:(,4.1.0]"
+  curations:
+    comment: |
```
fviernau, line 3: Can we shorten the comment please as follows?

"The LICENSE file always contained only the 'MIT' license text, while the project metadata accidentally  \
declared it to be license under 'BSD-LICENSE' for all version prior to 2.12.0, see also \
https://github.com/pytest-dev/pytest-cov/commit/2cd009442ddb08dc873117019b13ef972803d5c4"

curations/PyPI/_/pytest-cov.yml
```diff
@@ -0,0 +1,15 @@
+- id: "PyPI::pytest-cov:(,4.1.0]"
```
fviernau, line 1: commit-message: The rationale for the curation is supposed to be made clear by its comment. 
So, let's use this as a single source of truth and reduce the commit message to a title without body.
E.g.: 

chore: Map a declared license for `PyPI::pytest-cov`

fviernau (changes requested):

curations/PyPI/_/pytest-cov.yml
brunopacheco1: @fviernau thanks for your feedback. Can you please check again this PR?

curations/PyPI/_/pytest-cov.yml
```diff
@@ -0,0 +1,8 @@
...
+  curations:
+    comment: |
+      The LICENSE file always contained only the 'MIT' license text, while the project metadata accidentally
+      declared it to be license under 'BSD-LICENSE' for all version prior to 2.12.0, see also
```
sschuberth, line 5: s/license/licensed/
s/'BSD-LICENSE'/'BSD License'/

curations/PyPI/_/pytest-cov.yml
fviernau: Thanks for the changes, and sorry for the typos. See below.

curations/PyPI/_/pytest-cov.yml
sschuberth: > for all version prior to 2.12.0

If that's the case, why isn't the `id`'s version `(,2.12.0[`?

willebra:
@sschuberth I think this looks quite a genuine error at the project side, and I don't have doubts based on the history and the linked commits. The license-file has remained MIT all the time. And setup.py had internally conflicting statements in history, i.e. MIT and BSD License statements, both at the same time, prior to the correction. After correction, it is MIT in every place. 

Also any risk is quite low, considering the very similar obligations of these licenses, of course the BSD-ststement has been quite vague.

I could treat all the historical versions also as MIT.

curations/PyPI/_/pytest-cov.yml
brunopacheco1: No problem. Just fixed them.

curations/PyPI/_/pytest-cov.yml
brunopacheco1: Nice spot. Fixed.

curations/PyPI/_/pytest-cov.yml
```diff
@@ -0,0 +1,8 @@
+- id: "PyPI::pytest-cov:(,2.12.0["
+  curations:
+    comment: |
+      The LICENSE file always contained only the 'MIT' license text, while the project metadata accidentally
+      declared it to be licensed under 'BSD License' for all version prior to 2.12.0, see also
+      pytest-dev/pytest-cov@2cd0094
```
sschuberth, line 6: Please make this a full, clickable link. And better link the PR, which has some additional background information: https://github.com/pytest-dev/pytest-cov/pull/467

brunopacheco1: Done.

sschuberth: I was meaning to actually replace the link with *only* the link to the PR, which already contains the diff for the commit, so there is not need to link the commit as well.

fviernau (dismissed):

curations/PyPI/_/pytest-cov.yml
brunopacheco1: Ah, yes, no problem, just fixed the commit.

sschuberth (approved):

PR #160, opened 2023-12-21
fix(antlr): Map "BSD License" to `ANTLR-PD` for version 2.7.7

sschuberth:
Version 2.7.7 declares "BSD License" in its POM [1], which is wrong when looking at the `LICENSE.txt` contained in the source artifact. Instead, it is the ANTLR-specific Public Domain license, see [2]

[1]: https://repo1.maven.org/maven2/antlr/antlr/2.7.7/antlr-2.7.7.pom
[2]: https://spdx.org/licenses/ANTLR-PD.html

### DIFF
--- a/curations/Maven/antlr/antlr.yml
+++ b/curations/Maven/antlr/antlr.yml
@@ -21,3 +21,10 @@
       hash:
         value: "802655c343cc7806aaf1ec2177a0e663ff209de1"
         algorithm: "SHA1"
+- id: "Maven:antlr:antlr:2.7.7"
+  curations:
+    comment: |
+      Version 2.7.7 declares "BSD License" in its POM, which is wrong when looking at the `LICENSE.txt` contained in the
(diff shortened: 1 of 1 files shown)


sschuberth:
FYI @willebra.

curations/Maven/antlr/antlr.yml
```diff
@@ -16,6 +16,8 @@
   curations:
     comment: |
       The source distribution was found via the homepage.
+    declared_license_mapping:
+      BSD License: "ANTLR-PD"
```
mnonnenmacher, line 20: The explanation from the commit message should also be in the comment, as this is what will be contained in the ORT results and shown in reports. This is also a good candidate to make separate curations for the license mapping and the source artifact (as recently discussed in the core dev meeting) as they need different explanations.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #163, opened 2023-12-28
Add some categorizations for GCC exception 3.1

fviernau:
See individual commits.

### DIFF
--- a/license-classifications.yml
+++ b/license-classifications.yml
@@ -1082,6 +1082,11 @@ categorizations:
   - "copyleft-limited"
   - "include-in-notice-file"
   - "include-source-code-offer-in-notice-file"
+- id: "GPL-3.0-only WITH GCC-exception-3.1"
+  categories:
+  - "copyleft"
+  - "include-in-notice-file"
+  - "include-source-code-offer-in-notice-file"
 - id: "GPL-3.0-or-later"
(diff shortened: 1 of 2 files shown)


license-classifications.yml
```diff
@@ -1082,6 +1082,11 @@ categorizations:
   - "copyleft-limited"
   - "include-in-notice-file"
   - "include-source-code-offer-in-notice-file"
+- id: "GPL-3.0-only WITH GCC-exception-3.1"
```
sschuberth, line 1085: Just FYI, I was [thinking](https://github.com/doubleopen-project/policy-configuration/pull/64#issuecomment-1859758542) (with @willebra and others) about actually allowing to classify exceptions on their own. There's nothing built into ORT that currently prohibits that. And it would have the advantage that you don't need to classify e.g. `GCC-exception-3.1` with every license it can be combined with, but only `GCC-exception-3.1` on its own. Then, taking into account [ORT's exception mapping](https://github.com/oss-review-toolkit/ort/blob/main/utils/spdx/src/main/resources/exception-mapping.yml), you'd know the applicable licenses, and could handle the combination of licenses with exception only in rules, but not yet in the classification.

sschuberth (approved):

PR #167, opened 2024-02-08
chore(gradle): Upgrade Gradle wrapper to version 8.6

fviernau:


mnonnenmacher (dismissed):
This should also update the `gradlew.bat` script at least. I noticed that this often does not work if you run the wrapper task only once but if you run it again the wrapper scripts are updated.

sschuberth (approved):
I took the liberty to amend the commit with the necessary changes.

PR #169, opened 2024-02-19
feat(curations): Fixup SCM information for `eu.europa.ec.*` Maven projects

sschuberth:


curations/Maven/eu.europa.ec.bdmsl/bdmsl-api.yml
```diff
@@ -0,0 +1,7 @@
+- id: "Maven:eu.europa.ec.bdmsl:bdmsl-api"
```
fviernau, line 1: Should we say "Set" instead of "Fixup"?

To me it sounds a bit like existing curation has been fixed up.

fviernau: maybe say: Set VCS info for some `eu.europa.ec.bdmsl.*` artifacts

fviernau (approved):

PR #173, opened 2024-02-27
perf(evaluator.rules): Consistently prefer package- over dependency rule

fviernau:
For dependency trees made up of large amounts of package references relative to the amount of packages, `packageRule()` executes much faster compared to `dependencyRule()`. So, use `packageRule()` where possible to speed up the execution. For example, projects with large amounts of scopes, typical for Gradle based Android projects, can be evaluated faster.

### DIFF
--- a/evaluator.rules.kts
+++ b/evaluator.rules.kts
@@ -1109,7 +1109,7 @@ fun RuleSet.commercialInDependencyRule() = packageRule("COMMERCIAL_IN_DEPENDENCY
     }
 }
 
-fun RuleSet.copyleftInDependencyRule() = dependencyRule("COPYLEFT_IN_DEPENDENCY") {
+fun RuleSet.copyleftInDependencyRule() = packageRule("COPYLEFT_IN_DEPENDENCY") {
     require {
         -isProject()
         -isExcluded()
@@ -1354,7 +1354,7 @@ fun RuleSet.missingTestsRule() = projectSourceRule("MISSING_TESTS") {
     )
 }
 
-fun RuleSet.noLicenseInDependencyRule() = dependencyRule("NO_LICENSE_IN_DEPENDENCY") {
+fun RuleSet.noLicenseInDependencyRule() = packageRule("NO_LICENSE_IN_DEPENDENCY") {
     require {
         -isProject()
         -isExcluded()


evaluator.rules.kts
```diff
@@ -1109,7 +1109,7 @@ fun RuleSet.commercialInDependencyRule() = packageRule("COMMERCIAL_IN_DEPENDENCY
     }
 }
 
-fun RuleSet.copyleftInDependencyRule() = dependencyRule("COPYLEFT_IN_DEPENDENCY") {
+fun RuleSet.copyleftInDependencyRule() = packageRule("COPYLEFT_IN_DEPENDENCY") {
```
sschuberth, line 1112: Commit message: I'd propose `pref` instead of `fix` as a type.

sschuberth: How about renaming the rule and functions then as well, and use "package" instead of "dependency" as a term? I.e. `"COPYLEFT_IN_PACKAGE"` and `copyleftInPackageRule`.

fviernau: > How about renaming the rule and functions then as well, and use "package" instead of "dependency" as a term? I.e. `"COPYLEFT_IN_PACKAGE"` and `copyleftInPackageRule`.

I prefer not to, as dependency / package are similar terms. To me this is implementation detail and I wouldn't expose the user to two different terms.

evaluator.rules.kts
mnonnenmacher, line 1: Commit message: Performance is not the only difference. As the `DependencyRule` is executed for every node in the tree, it also means that issues are created for every time a package occurs in the tree. If the rule does not require a specific input of `DependencyRule` (ancestors, level, scope) this is also a reason to use `PackageRule` instead.

fviernau: > If the rule does not require a specific input of DependencyRule (ancestors, level, scope) this is also a reason to use PackageRule instead.

It does not require that specific input, which is why this is only a performance optimization without any impact on the result. E.g. the change is non-function.

Anyhow, do you want the commit message to be changed ? If so, what do you propose ?

mnonnenmacher: Yes, the proposal was to adapt the commit message to mention that this is not only a performance improvement but also a functional fix, as previously the rules would have created duplicate issues if a package occurred in the dependency tree more than once.

fviernau: IIRC the violations do get de-duplicated already since a while.

mnonnenmacher: Interesting, should this work globally? Because I just saw a report with an identical issue repeated a hundred times. 

Commit message nit: Typo in "Consistently"

fviernau: I've updated the commit message.

mnonnenmacher: Commit message:
* Typo in "Consistently" is still there
* Typo in "inserted"

mnonnenmacher (approved):

PR #175, opened 2024-04-03
fix(evaluator.rules): Re-align the `isException()` implementation

fviernau:
ORT nowadays reports stand-alone exceptions as license expressions of the form `'NOASSERTION WITH *'`.  Reflect that in `isException()`.


evaluator.rules.kts
```diff
@@ -966,7 +966,9 @@ fun resolveViolationInSourceCodeText(pkg: Package, license: String) : String {
  */
 
 fun isException(license: SpdxSingleLicenseExpression): Boolean =
-    ("-exception" in license.toString() && " WITH " !in license.toString())
+    license.toString().let { string ->
+        string.startsWith("NOASSERTION WITH ") ||  ("-exception" in string && " WITH " !in string)
+    }
```
sschuberth, line 971: I currently don't remember what was the original intent of this function. Judging from its use in this file, I guess the intent was to check for `license` to in fact be a *stand-alone* exception, and not a license *with* an exception (i.e. an `SpdxLicenseWithExceptionExpression`). If that's correct, the new code would change these semantics, because anything starting with "NOASSERTION WITH " *is* an `SpdxLicenseWithExceptionExpression`.

As the only use-case for this function seems to be the check whether a license is unhandled, I wonder whether we need this function (and related rule matchers) at all anymore. Because at least at Double Open we recently agreed with our legal people that [license classifications should explicitly include licenses with exceptions](https://github.com/doubleopen-project/policy-configuration/pull/64#issuecomment-1971093972), i.e. exceptions should not be classified on their own, but always in conjunction with their belonging license. If we could agree on the same semantics here, I believe we could remove this function altogether.

fviernau: > If that's correct, the new code would change these semantics, because anything starting with "NOASSERTION WITH " is an SpdxLicenseWithExceptionExpression.

If I understand you correctly, you say that any license containing a " WITH " operator is not a stand-alone exception, but a `SpdxLicenseWithExceptionExpression`. If so, what are you aiming at with this, why is this relevant for this discussion? ...I believe it works out fine here to say "NOASSERTION WITH "  represents a stand-alone exception, because the exception is not associated with any license identifier.

> As the only use-case for this function seems to be the check whether a license is unhandled, I wonder whether we need this function (and related rule matchers) at all anymore.

The intention of the `isException()` function usages has always been to ignore stand-alone exceptions. E.g. the policy rules do not handle stand-alone exceptions at all. This however stopped working once ORT started to associate stand-alone exceptions with `NOASSERTION`.

sschuberth: > If I understand you correctly, you say that any license containing a " WITH " operator is not a stand-alone exception, but a `SpdxLicenseWithExceptionExpression`.

Correct.

> If so, what are you aiming at with this, why is this relevant for this discussion?

As I tried to explain already, this is relevant background information to understand why your proposed code change also changes semantics: Instead of *never* returning `true` for `SpdxLicenseWithExceptionExpression` like before, after your change the function *may* return `true` for special `SpdxLicenseWithExceptionExpression`.

I'm not necessarily saying this is wrong (which is why I commented only instead of requesting changes), but I wanted to point it out for discussion, as the commit message did not mention this change of semantics, so I was unsure whether it's intended / you're aware of it.

> The intention of the `isException()` function usages has always been to ignore stand-alone exceptions.

Ok, then this probably is really just a matter of defining what "stand-alone exception", because to me "NOASSERTION WITH ..." is *not* a stand-alone exception (in the syntax sense), whereas to you it seems to be (in the semantic sense).

To solve this, I propose to add a follow-up commit that renames the function to e.g. `isExceptionWithoutLicense()` and / or some function documentation that clarifies on the intent of the function.

sschuberth (approved):

PR #176, opened 2024-04-03
chore: Re-generate license classifications

fviernau:
The updated content was generated by [1].

[1] ./gradlew generateLicenseClassifications

### DIFF
--- a/license-classifications.yml
+++ b/license-classifications.yml
@@ -89,6 +89,10 @@ categorizations:
   categories:
   - "permissive"
   - "include-in-notice-file"
+- id: "AML-glslang"
+  categories:
+  - "permissive"
+  - "include-in-notice-file"
 - id: "AMPAS"
   categories:
   - "permissive"
@@ -147,6 +151,10 @@ categorizations:
   - "permissive"
   - "include-in-notice-file"
 - id: "Adobe-2006"
+  categories:
+  - "permissive"
+  - "include-in-notice-file"
+- id: "Adobe-Display-PostScript"
   categories:
   - "permissive"
   - "include-in-notice-file"
@@ -220,6 +228,10 @@ categorizations:
   categories:
   - "permissive"
   - "include-in-notice-file"
+- id: "BSD-2-Clause-Darwin"
+  categories:
+  - "permissive"
+  - "include-in-notice-file"
 - id: "BSD-2-Clause-Patent"
   categories:
   - "permissive"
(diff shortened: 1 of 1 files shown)


license-classifications.yml
```diff
@@ -1328,12 +1409,11 @@ categorizations:
   - "copyleft-limited"
   - "include-in-notice-file"
   - "include-source-code-offer-in-notice-file"
-- id: "LLGPL"
+- id: "LOOP"
```
sschuberth, line 1412: Any idea why "LLGPL" is gone although it's still present at https://scancode-licensedb.aboutcode.org/llgpl.html?

fviernau: No. But I guess that it's been removed after the HTML has been generated. 
You can see generation date at the bottom of that HTML.

sschuberth: I've done the work for you and double-checked: While https://scancode-licensedb.aboutcode.org/index.json still contains the "LLGPL" entry, it apparently was [re-categorized as an exception](https://github.com/nexB/scancode-licensedb/commit/3743149e019d98d5aad8b349850a1e71f03b2e0c#diff-62b58a10810e0ddbfdb7541cde96bd1cb75ba3867774ec0f15fc9e7a25e4cadb), and thus [gets filtered out](https://github.com/oss-review-toolkit/ort-config/blob/b89250b37a29a7dd96e6d9b5e28c71c716e1b2f4/tools/curations/buildSrc/src/main/kotlin/ScanCodeLicenseDbClassifications.kt#L223). Please add that as an explanation to the commit message (this also explains the removal of "LicenseRef-scancode-llgpl").

What's left is the explanation / double-checking against as errors for the removal of "LicenseRef-scancode-us-govt-unlimited-rights", but I currently have no time to inspect that as well.

fviernau: Why do you think we should manually double check each removed entry?

sschuberth: To me, that's standard double-checking / quality assurance work that I also do for my own commits, just to be sure that there's no bug in the import task or elsewhere.

This falls into the category of never merging anything that you do not understand / can explain, which I believe is the way of working that the @oss-review-toolkit/core-devs  agreed on earlier.

fviernau: I've just re-run the import task, and then the `LicenseRef-scancode-us-govt-unlimited-rights` appreared again.

fviernau: > This falls into the category of never merging anything that you do not understand / can explain, which I believe is the way of working that the @oss-review-toolkit/core-devs agreed on earlier.

Well, that's double standard obviously.

sschuberth: > the `LicenseRef-scancode-us-govt-unlimited-rights` appreared again.

On the one hand this is good, as a removal that we couldn't explain was reverted. On the other hand it's bad, as apparently the import is non-deterministic, maybe due to some (temporary) IO issues not being handled properly... :confused:

PR #177, opened 2024-04-04
build: Upgrade Gradle to version 8.7

fviernau:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionSha256Sum=9631d53cf3e74bfa726893aee1f8994fee4e060c401335946dba2156f440f24c
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.6-bin.zip
+distributionSha256Sum=cb0da6751c2b753a16ac168bb354870ebb1e162e9083f116729cec9c781156b8
```
sschuberth, line 3: This should be 544c35d6bd849ae8a5ed0bcea39ba677dc40f49df7d1835561582da2009b961d.

gradlew.bat
```diff
@@ -43,11 +43,11 @@ set JAVA_EXE=java.exe
 %JAVA_EXE% -version >NUL 2>&1
 if %ERRORLEVEL% equ 0 goto execute
 
-echo. 1>&2
-echo ERROR: JAVA_HOME is not set and no 'java' command could be found in your PATH. 1>&2
-echo. 1>&2
-echo Please set the JAVA_HOME variable in your environment to match the 1>&2
-echo location of your Java installation. 1>&2
+echo.
```
sschuberth, line 46: Hmm. I wonder why we do not see these changes to `gradlew.bat` in [ORT's upgrade to Gradle 8.7](https://github.com/oss-review-toolkit/ort/pull/8448/files). Did you really run

    ./gradlew wrapper --gradle-version=8.7 --gradle-distribution-sha256-sum=544c35d6bd849ae8a5ed0bcea39ba677dc40f49df7d1835561582da2009b961d

*twice* in each directory to do the update?

fviernau: I wasn't 100% certain so I copy pasted your command and ran it again. Seem now it's fixed.

I believe that before I've done the update using `gradle` binary on the `PATH` instead of the local `./gradlew` for doing the upgrade. Could this have caused the difference?

sschuberth: Probably yes, as then the wrapper code itself does not use the latest version.

sschuberth (approved):

PR #186, opened 2024-04-23
refactor(tools): Migrate `VerifyPackageCurationsTask` to a Kotlin script

sschuberth:
This is a proof-of-concept migration to show how verification would work with stand-alone Kotlin scripts. Ultimately, the goal is to migrate all tools to stand-alone Kotlin scripts in order to get rid of the Gradle project that just serves as the infrastructure to run tools tasks.


sschuberth:
@oss-review-toolkit/kotlin-devs please give this a try, and if we'll all good with this direction, I'll migrate the other tools as well.

For editing the scripts, you should be able to just open the root Gradle project in Fleet / IntelliJ to get syntax-highlighting and auto-completion.

sschuberth:
For the record, `.main.kts` files to support imports (aka "includes"), see e.g. https://github.com/Kotlin/kotlin-script-examples/blob/5036d149f86ce7e485cb1cf61520e01812871595/jvm/jsr223/jsr223-main-kts/testData/import-middle.main.kts#L2.

PR #193, opened 2024-06-03
feat(rules): Add CVSS 3.1/4 to high vulnerability in dependency rule

tsteenbe:
Prior to this change a package with a high severity vulnerability encoded in CVSS 3.1 or 4.0 would not trigger a policy rule violation.


evaluator.rules.kts
```diff
@@ -1546,10 +1546,16 @@ fun RuleSet.vulnerabilityWithHighSeverityInDependencyRule() = packageRule("HIGH_
         -isProject()
         -isExcluded()
         +AnyOf(
-            hasVulnerability(maxAcceptedSeverity, "CVSS2") { value, threshold ->
+            hasVulnerability(maxAcceptedSeverity, "CVSS:2") { value, threshold ->
```
sschuberth, line 1549: Note that this would now disregard vulnerabilities that literally use "CVSS2" as the scoring system as we compare scoring systems by their strict string representation currently:

https://github.com/oss-review-toolkit/ort/blob/f560e02d555fd3b511d33b9d35cc4baa7de79ce8/evaluator/src/main/kotlin/PackageRule.kt#L98

Similar below. So this change likely has unwanted side effects.

sschuberth: I'll take a brief look at changing the `hasVulnerability` rule itself to address that.

PR #194, opened 2024-06-06
feat(curations): Add a VCS curation for XMLUnit < 2.0

sschuberth:
Note that XMLUnit 2.x is hosted at [1].

[1]: https://github.com/xmlunit/xmlunit

### DIFF
--- a/curations/Maven/xmlunit/xmlunit.yml
+++ b/curations/Maven/xmlunit/xmlunit.yml
@@ -1,7 +1,11 @@
 - id: "Maven:xmlunit:xmlunit:(,2.0["
   curations:
     comment: |-
-      Wrong SVN repository URL in POM: https://repo1.maven.org/maven2/xmlunit/xmlunit/1.3/xmlunit-1.3.pom
+      The POM, e.g. https://repo1.maven.org/maven2/xmlunit/xmlunit/1.3/xmlunit-1.3.pom, contains a `BSD-3-Clause`
+      license header but a non-SPDX license name tag. Also the SVN repository URL is wrong as it either points to the
+      view URL or hard-codes the trunk revision.
+    declared_license_mapping:
+      BSD License: "BSD-3-Clause"
     vcs:
       type: "Subversion"
       url: "https://svn.code.sf.net/p/xmlunit/code"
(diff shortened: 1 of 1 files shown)


curations/Maven/xmlunit/xmlunit.yml
```diff
@@ -0,0 +1,7 @@
+- id: "Maven:xmlunit:xmlunit:(,2.0["
```
fviernau, line 1: commit message: How about the following title? 

feat(curations): Set the VCS info for `xmlunit`

curations/Maven/xmlunit/xmlunit.yml
```diff
@@ -1,7 +1,11 @@
 - id: "Maven:xmlunit:xmlunit:(,2.0["
   curations:
     comment: |-
-      Wrong SVN repository URL in POM: https://repo1.maven.org/maven2/xmlunit/xmlunit/1.3/xmlunit-1.3.pom
+      The POM, e.g. https://repo1.maven.org/maven2/xmlunit/xmlunit/1.3/xmlunit-1.3.pom, contains a `BSD-3-Clause`
+      license header but a non-SPDX license name tag. Also the SVN repository URL is wrong as it either points to the
+      view URL or hard-codes the trunk revision.
+    declared_license_mapping:
```
fviernau, line 7: Same here:

feat(curations): Map a declared license for `xmlunit`

fviernau (changes requested):

curations/Maven/xmlunit/xmlunit.yml
sschuberth: I'd like to emphasize it's for versions < 2 only, as 2.x versions do not have these issues anymore.

curations/Maven/xmlunit/xmlunit.yml
sschuberth: I basically have no objections, but I'd just like to understand: How is that message better than mine?

fviernau: For the bottom commit "curation(s)" is not mentioned twice...less redundant.
The proposed way is a bit shorter and focus a tad more on describing the effect on the curation, not so much the property used for it.

fviernau (approved):

curations/Maven/xmlunit/xmlunit.yml
sschuberth: I see, thanks for explaining. I made the messages a bit more compact.

PR #207, opened 2024-07-09
Set the VCS revision for Dirkster NuGet packages

sschuberth:
FYI @haikoschol.

### DIFF
--- a/curations/NuGet/Dirkster/MLib.yml
+++ b/curations/NuGet/Dirkster/MLib.yml
@@ -1,0 +1,7 @@
+- id: "NuGet:Dirkster:MLib:1.3.1.2"
+  curations:
+    comment: |
+      Tag missing in Git, revision found by browsing release date history, see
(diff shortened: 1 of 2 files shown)


curations/NuGet/Dirkster/MLib.yml
```diff
@@ -0,0 +1,5 @@
+- id: "NuGet:Dirkster:MLib:1.3.1.2"
+  curations:
+    comment: "Tag missing in Git, revision found by browsing release date history."
```
fviernau, line 3: Can you please add link link which provides evidence?

fviernau (changes requested):

curations/NuGet/Dirkster/MLib.yml
sschuberth: Added to the commit message.

fviernau: I believe that this should be part of the curation comment, because a curation should be self-contained.

fviernau (changes requested):

fviernau (approved):

PR #210, opened 2024-07-09
Set the source artifact for DynamicDataDisplay NuGet package

haikoschol:


curations/NuGet/_/DynamicDataDisplay.yml
```diff
@@ -0,0 +1,10 @@
+- id: "NuGet::DynamicDataDisplay:0.3.0"
+  curations:
+    comment: |
+      Repo linked in metadata was deleted. The source artifact was created from
+      the only commit added on the date version 0.3.0 was published.
```
sschuberth, line 5: I guess we should explain a bit more how the archive was created. I propose:

> The repository linked in the package's metadata was deleted. However, it is still accessible at Software Heritage, see https://archive.softwareheritage.org/browse/origin/directory/?origin_url=https://github.com/apoydence/DynamicDataDisplay. From there, an archive was created from the only commit added on the date when version 0.3.0 was released.

sschuberth (approved):

PR #222, opened 2024-07-18
feat(curations): Set the VCS info for "PyPI::tensorflow-intel"

sschuberth:
FYI @haikoschol, this also has the https://github.com/Intel-tensorflow/tensorflow/tree/v2.12.0 tag.

### DIFF
--- a/curations/PyPI/_/tensorflow-intel.yml
+++ b/curations/PyPI/_/tensorflow-intel.yml
@@ -1,0 +1,5 @@
+- id: "PyPI::tensorflow-intel:"
+  curations:
+    comment: No VCS information contained in metadata. The correct VCS URL was found via Internet search.
+    vcs:
+      url: "https://github.com/Intel-tensorflow/tensorflow.git"


haikoschol:
> FYI @haikoschol, this also has the https://github.com/Intel-tensorflow/tensorflow/tree/v2.12.0 tag.

This looks like a fork of Tensorflow though, not the "Intel extensions for Tensorflow". IIUC, the latter allows running computation in Tensorflow on Intel hardware with some optimisations.

curations/PyPI/_/tensorflow-intel.yml
```diff
@@ -0,0 +1,5 @@
+- id: "PyPI::tensorflow-intel:"
+  curations:
+    comment: The correct VCS URL was found via Internet search.
```
fviernau, line 3: For setting the VCS info?
1. Should the comment make clear whether a) VCS info was present but wrong b) was missing
2. Should the curations the the VCS type as well? (even though maybe not strictly needed)

fviernau (changes requested):

curations/PyPI/_/tensorflow-intel.yml
sschuberth: > Should the comment make clear whether a) VCS info was present but wrong b) was missing

I've added an according sentence. But just curious: What's the use-case of documenting this?

> Should the curations the the VCS type as well?

I'm unopinionated here. But I actually believe that you earlier preferred to only curate what cannot be determined automatically.

sschuberth:
> This looks like a fork of Tensorflow though, not the "Intel extensions for Tensorflow". IIUC, the latter allows running computation in Tensorflow on Intel hardware with some optimisations.

It's the same for Intel's "friendly fork" I believe, see the [release notes from that repository](https://github.com/Intel-tensorflow/tensorflow/releases/tag/v2.12.0):

> This release of Intel® Optimized TensorFlow is based on the TensorFlow [v2.12.0 tag](https://github.com/tensorflow/tensorflow/releases/tag/v2.12.0) and is built with support for oneDNN (oneAPI Deep Neural Network Library).

fviernau (approved):

curations/PyPI/_/tensorflow-intel.yml
fviernau: > I've added an according sentence. But just curious: What's the use-case of documenting this?

For the reviewer to understand the need. But also for the maintaining the curation, e.g. why was it invented. 
There is no information for which version is the package the curation was created, so it would not be possible to figure out the reason later on anymore. This may be interresting in case of a future needed update, e.g. package starts having VCS info and in particular if repository is moved.

fviernau: > I'm unopinionated here. But I actually believe that you earlier preferred to only curate what cannot be determined automatically.

The curations I remember (without double checking), always set VCS type and url if no VCS info in pkg metadata.
Also the ones I've created IIRC followed this pattern. So, not following here.

sschuberth: > For the reviewer to understand the need.

Again, just out of curiosity: In what regard is the need different for non-existing or wrong VCS info that needs to be curated?

> The curations I remember (without double checking), always set VCS type and url if no VCS info in pkg metadata.

Could be. In any case, we should implement whatever we prefer as PR checks (also see https://github.com/oss-review-toolkit/ort-config/pull/186) to avoid discussions.

PR #250, opened 2024-11-06
curations: Update curations for Nuget packages

mnonnenmacher:
Update the curations by running the generation tasks in `tools/curations`.

### DIFF
--- a/curations/NuGet/Microsoft.AspNetCore/Http.yml
+++ b/curations/NuGet/Microsoft.AspNetCore/Http.yml
@@ -1,18 +1,23 @@
-- id: "NuGet:Microsoft.AspNetCore:Http:(,2.1.22["
-  curations:
-    comment: "Set the VCS path of the module inside the multi-module repository."
-    vcs:
-      path: "src/Microsoft.AspNetCore.Http"
-- id: "NuGet:Microsoft.AspNetCore:Http:[2.1.22,2.1.34]"
-  curations:
-    comment: "Set the VCS path of the module inside the multi-module repository."
-    vcs:
-      path: "src/Http/Http/src"
 - id: "NuGet:Microsoft.AspNetCore:Http:(2.1.34,3.0.0["
   curations:
     comment: "Set the VCS path of the module inside the multi-module repository."
     vcs:
       path: "src/Microsoft.AspNetCore.Http"
+- id: "NuGet:Microsoft.AspNetCore:Http:(,2.1.22["
+  curations:
+    comment: "Set the VCS path of the module inside the multi-module repository."
+    vcs:
+      path: "src/Microsoft.AspNetCore.Http"
+- id: "NuGet:Microsoft.AspNetCore:Http:[2.1.7,)"
+  curations:
+    comment: "Set the VCS path of the module inside the multi-module repository."
+    vcs:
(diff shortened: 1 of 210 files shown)


nnobelis (approved):

sschuberth:
IIRC, @fviernau did some corrections to these generated curation in https://github.com/oss-review-toolkit/ort-config/pull/225. I hope these do not get overwritten now.

mnonnenmacher:
> IIRC, @fviernau did some corrections to these generated curation in #225. I hope these do not get overwritten now.

That's really unfortunate, why fix the generated code instead of the generator? Also, it appears to me that the "fixes" did not fix any real issues as curations for non-published packages do not do any harm except for wasting a little space.
I won't have time to go through the >200 files anytime soon.

sschuberth:
> curations for non-published packages do not do any harm

Then maybe I'm mistaken, as I recall during the review that (some of) these packages *are* published. @fviernau can probably tell better.

mnonnenmacher:
> > IIRC, @fviernau did some corrections to these generated curation in #225. I hope these do not get overwritten now.
> 
> That's really unfortunate, why fix the generated code instead of the generator? Also, it appears to me that the "fixes" did not fix any real issues as curations for non-published packages do not do any harm except for wasting a little space. I won't have time to go through the >200 files anytime soon.

I had a look at one single package (`NuGet:Microsoft.AspNetCore:Authentication`) and it appears there is some problem with the generator. However, I still don't understand why this was not addressed by fixing the generator, the manual fixes probably missed many packages?

fviernau:
@mnonnenmacher I have a hunch you do not really understand the fixes I've made, so
why judging about the approach without asking the question why? e.g.: 

> That's really unfortunate, why fix the generated code instead of the generator?

The effort it took to investigate the issues caused by all these wrong curations was that huge,
that I would have probably been faster if they were not existent at all.  

> Also, it appears to me that the "fixes" did not fix any real issues as curations for non-published packages do not do any harm except for wasting a little space.

Each of these fixes did fix any real issues, in contrast to the generated curations, where many of them seems to
have been introduced for non-real issues.

>  However, I still don't understand why this was not addressed by fixing the generator, the manual fixes probably missed many packages?

Because the approach how these are generated does not work, and coming up with a better approach would mean redo it from scratch.

curations/NuGet/Microsoft.AspNetCore/Http.yml
```diff
@@ -1,18 +1,23 @@
...
     vcs:
       path: "src/Microsoft.AspNetCore.Http"
-- id: "NuGet:Microsoft.AspNetCore:Http:[2.1.22,2.1.34]"
+- id: "NuGet:Microsoft.AspNetCore:Http:[2.1.7,)"
   curations:
     comment: "Set the VCS path of the module inside the multi-module repository."
     vcs:
       path: "src/Http/Http/src"
-- id: "NuGet:Microsoft.AspNetCore:Http:(2.1.34,3.0.0["
+- id: "NuGet:Microsoft.AspNetCore:Http:[2.1.22,2.1.34]"
```
fviernau, line 16: Please do not revert my fixes.

fviernau (changes requested):

PR #260, opened 2025-01-16
feat(curations): Add a VCS curation for "Maven:net.sf.saxon:saxon:9"

sschuberth:


tsteenbe (commented):
Where did you find version 9? On [Maven Central](https://central.sonatype.com/artifact/net.sf.saxon/saxon/versions) I only see 8.7 as the latest released version for this Saxon namespace.

Seeing you state "The POM does not provide an SCM tag" plus version 9 is not Maven Central  I would like to see in the comment a link to the pom.xml so it easy for anyone to verify the correctness of this curation.

fviernau (approved):

PR #263, opened 2025-01-18
More saxon curations

sschuberth:


tsteenbe (changes requested):
Understand why you would extract base properties for Saxon but the id/comment combination is correct as https://repo1.maven.org/maven2/net/sf/saxon/saxon/8.7/saxon-8.7.pom does include SCM tag. Recommend to update the comment to use Maven Central pom location but mention SCM url does not work anymore.

tsteenbe (changes requested):
Would like to request 2 more changes:

1.  Drop the `comment` for id: `Maven:net.sf.saxon:saxon:9` curation it mentionsthe POM in vcs revision curation - makes no sense to me.
2.  Improve the `comment` for `Maven:net.sf.saxon:saxon` curation  to include POM links for 8.7 and 9 to as proof of correctness of curation. 

Maybe change from:

```
- id: "Maven:net.sf.saxon:saxon"
  curations:
    comment: |-
      The POMs for the different versions either have no or an outdated SCM tag.
```

to

```
- id: "Maven:net.sf.saxon:saxon"
  curations:
    comment: |-
      The POMs for the different versions either have no or an outdated SCM tag.
      See https://repo1.maven.org/maven2/net/sf/saxon/saxon/8.7/ and
      https://nexus.intranda.com/repository/maven-releases/net/sf/saxon/saxon/9/saxon-9.pom 
```

-

sschuberth:
> 1. it mentionsthe POM in vcs revision curation - makes no sense to me.

Why not? If the POM contained an `scm` tag, then no VCS curation would be needed. Also in the initial version [you explicit requested the link to the POM to be added](https://github.com/oss-review-toolkit/ort-config/pull/260#pullrequestreview-2556623038).

> curation to include POM links for 8.7 and 9 to as proof of correctness of curation.

There are more version that those, and I don't think it makes sense to add each of them as a proof. So I'll pick just one as an example.

sschuberth:
Ok, I now get your point: You're arguing that now with the https://nexus.intranda.com/repository/maven-releases/net/sf/saxon/saxon/9/saxon-9.pom link being part of the more generic curation, it does not need to be repeated in the version-specific curation. I agree to that.

PR #288, opened 2025-04-16
feat(evaluator.rules): Improve the unmapped declared license rule

fviernau:
A concluded license overrides the applicable license including any declared license. So, do not report issues about the declared license if a concluded license is set.

### DIFF
--- a/evaluator.rules.kts
+++ b/evaluator.rules.kts
@@ -1601,6 +1601,7 @@ fun RuleSet.unhandledLicenseRule() = packageRule("UNHANDLED_LICENSE") {
 fun RuleSet.unmappedDeclaredLicenseRule() = packageRule("UNMAPPED_DECLARED_LICENSE") {
     require {
         -isExcluded()
+        -hasConcludedLicense()
     }
 
     resolvedLicenseInfo.licenseInfo.declaredLicenseInfo.processed.unmapped.forEach { unmappedLicense ->


evaluator.rules.kts
```diff
@@ -1601,6 +1601,7 @@ fun RuleSet.unhandledLicenseRule() = packageRule("UNHANDLED_LICENSE") {
 fun RuleSet.unmappedDeclaredLicenseRule() = packageRule("UNMAPPED_DECLARED_LICENSE") {
     require {
         -isExcluded()
+        -hasConcludedLicense()
```
sschuberth, line 1604: Shouldn't then also `howToFixUnmappedDeclaredLicense()` be changed to include setting a concluded license as an alternative way to fix this violation?

fviernau: While setting the concluded license would fix the issue, I would prefer to not recommend it as a solution as I believe it's suboptimal and users who choose to adress this issue with a concluded license should already have a good reason anyway for doing so.

sschuberth: Fine with me, but I'd also like @tsteenbe to confirm, as I'm not making use of these rules anyway.

tsteenbe: @fviernau  Would be fine with this change in principal but just wondering the impact on generated output files. I am correct to say unmapped declared licenses are never included in ORT generated SBOM and notices?

fviernau: > I am correct to say unmapped declared licenses are never included in ORT generated SBOM and notices?

That's correct AFAIK.

tsteenbe: Than I am OK with this change

tsteenbe (approved):

PR #450, opened 2026-07-09
feat(text-templates): Add third-party-notices template

tsteenbe:
New template to help user more easily create the notices that they want/their org is asking from to see output see 
[THIRD-PARTY-NOTICES.txt](https://github.com/user-attachments/files/29833948/THIRD-PARTY-NOTICES.txt)

This PR introduces a new template to help users create third-party notices more easily, based ob common  requirements learned from various organizations in ORT's community.

**Example Output**

To see example output from the template, see [THIRD-PARTY-NOTICES.txt](https://github.com/user-attachments/files/29833948/THIRD-PARTY-NOTICES.txt)

**Testing**

To test this new template, download and extract the test project in  [notices-generation-test-project.zip](https://github.com/user-attachments/files/29837680/notices-generation-test-project.zip) and follow instructions in README.md.

### DIFF
--- a/text-templates/THIRD-PARTY-NOTICES.txt.ftl
+++ b/text-templates/THIRD-PARTY-NOTICES.txt.ftl
@@ -1,0 +1,266 @@
+[#--
+    Copyright (C) 2020 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort-config/blob/main/NOTICE>)
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+    SPDX-License-Identifier: Apache-2.0
+    License-Filename: LICENSE
+--]
+[#--
+    This template produces a third-party attribution notices document listing the SPDX license expressions,
+    license texts and copyright statements of the analyzed projects and their dependencies. The
+    output is plain text that uses reStructuredText section markup (titles underlined with '=' for
+    the document, '-' for packages and '~' for the blocks within a package), so it can be read as-is
+    or rendered to HTML/PDF with a reStructuredText processor such as docutils.
+
+    Only licenses categorized as "include-in-notice-file" are shown. Excluded projects and packages,
+    and packages without any notice-worthy license, are omitted. The document is structured as:
+
+    1. A "Third-Party Notices" title.
+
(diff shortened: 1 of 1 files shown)


text-templates/THIRD-PARTY-NOTICES.txt.ftl
```diff
@@ -0,0 +1,233 @@
+[#--
+    Copyright (C) 2020 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort-config/blob/main/NOTICE>)
```
sschuberth, line 2: Why is this 2020? Wasn't this written from scratch?

sschuberth (commented):
As a general remark, it slipped my mind that we now have a "text-templates" directory that _also_ contains a "NOTICE_SUMMARY.ftl" like the one we have [here](https://github.com/oss-review-toolkit/ort/blob/main/plugins/reporters/freemarker/src/main/resources/templates/plain-text/NOTICE_SUMMARY.ftl). I believe we should make clearer to users which one to prefer by only using either location.

text-templates/THIRD-PARTY-NOTICES.txt.ftl
tsteenbe: @sschuberth It's contains bits of NOTICE_BY_PACKAGE code which is why I choose to keep 2020 copyrights.

tsteenbe:
> As a general remark, it slipped my mind that we now have a "text-templates" directory that _also_ contains a "NOTICE_SUMMARY.ftl" like the one we have [here](https://github.com/oss-review-toolkit/ort/blob/main/plugins/reporters/freemarker/src/main/resources/templates/plain-text/NOTICE_SUMMARY.ftl). I believe we should make clearer to users which one to prefer by only using either location.

The FTL files in ORT are  for me only for testing - the ones users should use as reference are in ort-config. Have some follow up PRs coming to ort and ort-config after which I think it will become much clearer for users - includes rewrite of docs, more config files including FTL files.

sschuberth:
> The FTL files in ORT are for me only for testing

That's new to me. If we'd want these to be only for testing, they should be _test_ resources instead of _main_ resources, and we'd not need the mechanism to [specify templates by resources ID](https://github.com/oss-review-toolkit/ort/blob/183e8fd110b18aa1aeca4baf7ec0979997d2a404/plugins/reporters/asciidoc/src/main/kotlin/AsciiDocTemplateReporter.kt#L37-L43).

tsteenbe:
> That's new to me. If we'd want these to be only for testing, they should be test resources instead of main resources, and we'd not need the mechanism to [specify templates by resources ID](https://github.com/oss-review-toolkit/ort/blob/183e8fd110b18aa1aeca4baf7ec0979997d2a404/plugins/reporters/asciidoc/src/main/kotlin/AsciiDocTemplateReporter.kt#L37-L43).

I am saying they only used for testing because every org I worked with on this topic want to make changes so the notices meets their specific needs. In this PR I merge the wishes of several of then into one FTL file and putting it here so can easily solicit feedback and use it in ORT CI integrations.

